PR #35, opened 2017-07-02
admit: Fix tempfile security (fixes #4)

### LINKED ISSUE #4: Improve tempfile security of admit.py
It does sketchy things.

andersk:
ssh-keygen makes up its own filename without taking special precautions about being in a shared tmp directory, so do it in our own temporary directory instead.

Fixes #4.


celskeggs:
Would it be possible for you to sign this commit?

PR #37, opened 2017-07-02
Improve shell quoting hygiene

andersk:


celskeggs:
Looks pretty good. Can you fix the conflicts?

PR #68, opened 2017-08-10
Refuse to sign wildcard ssh certs

ichung:
Add an extra check in the SSH authority to prevent signing certificates with 0 valid principals, since such certificates are treated as valid for all principals.  (We already check this during config compilation, but I consider it important enough to check again.)

### DIFF
--- a/building/build-debs/homeworld-keyserver/src/account/privileges.go
+++ b/building/build-debs/homeworld-keyserver/src/account/privileges.go
@@ -1,9 +1,10 @@
 package account
 
 import (
-	"authorities"
 	"fmt"
 	"time"
+
+	"authorities"
 	"token"
 )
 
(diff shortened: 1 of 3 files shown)


building/build-debs/homeworld-keyserver/src/account/privileges.go
```diff
@@ -57,7 +58,7 @@ func stringInList(value string, within []string) bool {
 }
 
 func NewBootstrapPrivilege(allowed_principals []string, lifespan time.Duration, registry *token.TokenRegistry) (Privilege, error) {
-	if allowed_principals == nil || len(allowed_principals) == 0 || lifespan < time.Millisecond || registry == nil {
+	if len(allowed_principals) == 0 || lifespan < time.Millisecond || registry == nil {
```
ichung: They're redundant since `len` of a nil slice is 0.

PR #74, opened 2017-09-15
Merge key management infrastructure code

celskeggs:
This change removes the old admission and authorization code and adds new key management code.

The new keysystem isn't fully integrated yet, so this PR is not yet ready to merge, but it would be useful to start getting reviews on this beyond the existing reviews that have happened of parts of the keysystem code.


celskeggs:
Reminder to self: I have a few TODOs in buglog.txt, which should be resolved before merging.

celskeggs:
I'm continuing my work in the feature-ki-deployable branch for the sake of not giving readers even more of a mess to review.

celskeggs:
@ichung: when you get a chance, this is on the critical path, since DC4 is now deployed and tested. There are also more commits on `feature-ki-deployable`, but this should be reviewed and merged first.

building/build-debs/homeworld-keysystem/src/keyclient/actloop/actloop_test.go
```diff
@@ -0,0 +1,387 @@
+package actloop
```
ichung: Some questions about how `keyclient` works in general:
- Why is so much state stored on disk, rather than in the `state` struct?
- Doesn't the keygranting cert expire?  I don't see where we renew it.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	return ServerEndpoint{rootCAs: authorities, baseURL: url, timeout: time.Second * 30}, nil
+}
+
+func (s ServerEndpoint) BaseURL() string {
+	return s.baseURL
+}
+
+func (s ServerEndpoint) WithHeader(key string, value string) ServerEndpoint {
+	out := s
```
ichung: The function call already makes a copy.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
```diff
@@ -0,0 +1,108 @@
...
+	return out
+}
+
+func (s ServerEndpoint) WithCertificate(cert tls.Certificate) ServerEndpoint {
+	out := s
+	out.certificates = make([]tls.Certificate, len(s.certificates))
+	copy(out.certificates, s.certificates)
+	out.certificates = append(out.certificates, cert)
+	return out
+}
```
ichung: I think this entire function body could be replaced with

    s.certificates = append(s.certificates, cert)
    return s

(Unless it is necessary that the new `certificates` slice does not share space with the old one.  But I don't see any reason why that should be so.)

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
```diff
@@ -0,0 +1,108 @@
...
+func (s ServerEndpoint) Request(path string, method string, reqbody []byte) ([]byte, error) {
+	if path[0] != '/' {
+		return nil, errors.New("while validating request: path must be absolute")
+	}
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs:      s.rootCAs,
+				Certificates: s.certificates,
+			},
```
ichung: May as well set `MinVersion: tls.VersionTLS12`.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+		Timeout: s.timeout,
```
ichung: Should disable HTTP compression if partial-chosen-plaintext attacks are a possibility, a la [BREACH](https://en.wikipedia.org/wiki/BREACH).  Probably also worth looking at any other places we use HTTP over TLS (or other protocols) and see if they have compression on too.

building/build-debs/homeworld-keysystem/src/keycommon/knc/knc.go
```diff
@@ -0,0 +1,58 @@
...
+	"encoding/json"
+	"keycommon/reqtarget"
+)
+
+type KncServer struct {
+	Hostname string
+}
+
+func (k KncServer) kncRequest(data []byte) ([]byte, error) {
+	cmd := exec.Command("/usr/bin/knc", fmt.Sprintf("host@%s", k.Hostname), "20575")
```
ichung: The service name is always `host`?

building/build-debs/homeworld-keysystem/src/keytool.go
```diff
@@ -0,0 +1 @@
+package main
```
ichung: What is this here for?  In fact, why do we have 4 incompatible definitions of a single `main` package at the top level?  Shouldn't e.g. `keyserver.go` be in a subdirectory of `keyserver`, etc?

building/build-debs/homeworld-keysystem/src/util/fileutil/util.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return err // hard to test in practice, but still necessary
+		}
+	}
+	if !fileinfo.IsDir() {
+		return errors.New("not a directory")
+	}
+	return nil
+}
+
```
ichung: Anything using these functions suffers from TOCTTOU.  It's not a problem if we always have our own isolated filesystem, I guess?

building/build-debs/homeworld-keysystem/src/util/fileutil/util.go
```diff
@@ -0,0 +1,57 @@
...
+	} else {
+		return false
+	}
+}
+
+func EnsureIsFolder(dirname string) error {
+	fileinfo, err := os.Stat(dirname)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = os.MkdirAll(dirname, 0755)
```
ichung: Why so permissive?

building/build-debs/homeworld-keysystem/src/util/netutil/parseip.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, err
+	}
+	var ip net.IP
+	if remote_addr[0] == '[' && remote_addr[index-1] == ']' {
+		ip = net.ParseIP(remote_addr[1 : index-1])
+		if ip == nil {
+			return nil, fmt.Errorf("Invalid request address (invalid IP in '%s')", remote_addr)
+		}
+		if ip.To4() != nil {
+			return nil, fmt.Errorf("IP address was expected to be IPv6, but was IPv4")
```
ichung: What about, for instance, `[::ffff:0808:0808]` which is the IPv6-mapped version of the IPv4 address `8.8.8.8`?

building/build-debs/homeworld-keysystem/config/keyserver.yaml
```diff
@@ -0,0 +1,268 @@
...
+    group: root-admins
+    privilege: bootstrap-account
+    scope: nodes
+    lifespan: 1h
+
+  renew-keygrant:
+    group: nodes
+    privilege: sign-tls
+    authority: keygranting
+    lifespan: 960h  # fourty day lifespan
```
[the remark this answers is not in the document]
celskeggs: Since this one is pretty minor, and has already been fixed in the future commit d81ebe69aca170a9ef87d1ee76b77322af6a3d6d on feature-admin-tools, I'm skipping dealing with it now.

building/build-debs/homeworld-keysystem/src/keyclient/actions/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,113 @@
...
+	"util/fileutil"
+)
+
+type BootstrapAction struct {
+	State         *state.ClientState
+	TokenFilePath string
+	TokenAPI      string
+}
+
+func PrepareBootstrapAction(s *state.ClientState, tokenfilepath string, api string) (actloop.Action, error) {
```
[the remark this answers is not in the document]
celskeggs: Having a more consistent and encapsulated interface across the actions?

building/build-debs/homeworld-keysystem/src/keyclient/actloop/actloop_test.go
celskeggs: What do you mean about "so much state" being stored on disk? Keys and certs are stored on disk, as are downloaded files, because that's where downstream consumers pull them from. The only thing I can think of that you might be thinking about would be the bootstrap token, which needs to be on disk so that it's preserved across restarts of the service.

The keygranting cert is renewed by having an entry in the `keys` section to handle renewing it. Something like this:

      - name: keygranting
        type: tls
        key: /etc/hyades/keyclient/granting.key
        cert: /etc/hyades/keyclient/granting.pem
        api: renew-keygrant
        in-advance: 336h   # renew two weeks before expiration

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
celskeggs: It is necessary -- you could call two different `WithCertificate` calls on the same endpoint, and they should produce endpoints that don't conflict.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
celskeggs: Woops, yeah.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
celskeggs: +1

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
celskeggs: +1

building/build-debs/homeworld-keysystem/src/keycommon/knc/knc.go
celskeggs: Yes. Because our keytabs are for `host/egg-sandwich.mit.edu` and such.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
celskeggs: It looks like there aren't any other places.

building/build-debs/homeworld-keysystem/src/keyclient/actloop/actloop_test.go
ichung: I am still a little confused.  Where in the key renewal process does the `Keygrant` field get written to, so that we start using the new keygranting cert instead of the old one?

building/build-debs/homeworld-keysystem/src/keytool.go
celskeggs: These are in `main` because they need to be compiled into binaries? I'm not fully clear on how this system worked, but I can't just make these a subpackage of, say, keyserver, as far as I can tell. Go will just not produce a binary.

building/build-debs/homeworld-keysystem/src/util/fileutil/util.go
celskeggs: Because this is used to create directories like /etc/homeworld/authorities/ that have contents that need to be read by other users? So we restrict access to particular private keys as necessary, but otherwise let everything be readable by world, since they aren't security-critical.

building/build-debs/homeworld-keysystem/src/util/fileutil/util.go
celskeggs: Nothing in these functions is designed for security -- the files accessed are generally under places like /etc/homeworld/, and anyone with write access there can already gain full access to everything the keyclient has access to, because they can just read the private keys it has.

building/build-debs/homeworld-keysystem/src/keycommon/endpoint/endpoint.go
ichung: Oh, I see.  (Still no need for `out`.)

building/build-debs/homeworld-keysystem/src/keytool.go
ichung: You can put `keyserver.go` into `keyserver/main/`.

building/build-debs/homeworld-keysystem/src/util/netutil/parseip.go
celskeggs: Hmm, yeah.

building/build-debs/homeworld-keysystem/src/keyclient/actloop/actloop_test.go
celskeggs: Ah, woops -- I apparently dropped the call to ReloadKeygrantingCert from RequestOrRenewAction.Perform() during refactoring.

building/build-debs/homeworld-keysystem/src/keytool.go
celskeggs: Ahh, alright.

PR #106, opened 2017-10-07
Add documentation for verifying dependency signatures

krawthekrow:
To conserve hair.

### DIFF
--- a/how-to-verify-keys.md
+++ b/how-to-verify-keys.md
@@ -1,0 +1,34 @@
+Getting cela's Public Key
+========================
+
+1. Get an _untrusted_ copy of cela's key signature from the Repository Security section in homeworld's `README.md`.
+2. Get an _untrusted_ copy of cela's key from the MIT keyserver: `gpg --keyserver pgp.mit.edu --recv-keys <last 8 digits of untrusted cela's key signature>`
+3. Compute the fingerprint of the received key: `gpg --fingerprint cela`
+4. Ask cela for her fingerprint and check against the output of the previous command.
+5. Set the trust on cela's key with `gpg --edit-key cela`, `trust`, `4`, `quit` (`5` is reserved for your own key).
+
+Getting a Trusted Copy of `trust.txt`
+=====================================
+
+1. Clone the homeworld repository.
+2. Get the signatures of the commits for `trust.txt`: `git log --show-signature trust.txt`
+3. Check that the latest commit in the log is "good".
+
+Verifying a Dependency's Signature
+==================================
+
+1. Download an _untrusted_ copy of the dependency's signing key (e.g. `dependency_pubkey.gpg`).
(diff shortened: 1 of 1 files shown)


how-to-verify-keys.md
```diff
@@ -0,0 +1,34 @@
+Getting cela's Public Key
+========================
+
+1. Get an _untrusted_ copy of cela's key signature from the Repository Security section in homeworld's `README.md`.
+2. Get an _untrusted_ copy of cela's key from the MIT keyserver: `gpg --keyserver pgp.mit.edu --recv-keys <last 8 digits of untrusted cela's key signature>`
```
celskeggs: You may as well embed the last eight digits here.

how-to-verify-keys.md
```diff
@@ -0,0 +1,34 @@
...
+3. Compute the fingerprint of the received key: `gpg --fingerprint cela`
+4. Ask cela for her fingerprint and check against the output of the previous command.
+5. Set the trust on cela's key with `gpg --edit-key cela`, `trust`, `4`, `quit` (`5` is reserved for your own key).
```
celskeggs: What's different if you don't do this step?

how-to-verify-keys.md
```diff
@@ -0,0 +1,34 @@
...
+
+Getting a Trusted Copy of `trust.txt`
+=====================================
+
+1. Clone the homeworld repository.
+2. Get the signatures of the commits for `trust.txt`: `git log --show-signature trust.txt`
+3. Check that the latest commit in the log is "good".
```
celskeggs: Perhaps you want to have different criteria: not that the latest commit is "good", but checking who it says it's signed by? It seems like you might end up having previously trusted some other person's key, and then if they tried to sign these commits, it would also show up as "good", even though they aren't supposed to be signing these commits.

how-to-verify-keys.md
```diff
@@ -0,0 +1,34 @@
...
+
+Verifying a Dependency's Signature
+==================================
+
+1. Download an _untrusted_ copy of the dependency's signing key (e.g. `dependency_pubkey.gpg`).
+2. Import the dependency's signing key: `gpg --import <dependency_pubkey.gpg>`
```
celskeggs: I'm not sure we should be recommending importing the untrusted copy of the key, but I'm not quite sure how else to recommend to do this.

how-to-verify-keys.md
```diff
@@ -0,0 +1,32 @@
+Getting cela's Public Key
+========================
+
+1. Get an _untrusted_ copy of cela's key from the MIT keyserver: `gpg --keyserver pgp.mit.edu --recv-keys 0f825b91` (last 8 digits of cela's key signature)
```
celskeggs: Fingerprint, not signature.

PR #108, opened 2017-10-07
Updated build files, symlinks, checksums for go update to 1.8.4; made build files nicer

Lchu613:


celskeggs:
I think this was accidentally closed by the deletion of the devel-latest branch, but I'm not sure. I'm reopening it, because I have no idea why else it would be closed.

celskeggs:
If you're still up to work on this, can you rebase this against master?

Lchu613:
Okay, I'm still clumsy at git but I think that's what I just did.

celskeggs:
No, that rebase is thoroughly botched. The only commit that should be present is a1b6d6a.

Lchu613:
Oops. I'll try to unbotch that, then.

Lchu613:
That should be better, I think.
Also forgot to mention in the notes that this corresponds to pull request #4 in homeworld-upstream for source files.

celskeggs:
I can't verify your hash for the go 1.8.4 release -- it matches the package in your PR to homeworld-upstream, but that doesn't match what I can get from upstream myself. How did you retrieve this source package?

Relatedly, 1.8.5 is out, so it might be worthwhile to upgrade directly, rather than going through 1.8.4.

Lchu613:
Downloaded source package from golang website, untarred from .tar.gz, tarred to .tar.xz. Is there a better way to go about that?

celskeggs:
You can gunzip the tarball and re-xz it, without fully extracting the tarball.

celskeggs:
I'm assuming at this point that this PR is unlikely to be completed, and will close it and free up the issue for someone else to take a stab at. If I'm wrong, I'd be happy to reopen the PR.

PR #112, opened 2017-10-10
Upgrade etcd to 3.2.9

krawthekrow:
Do I need to update debian/changelog too? I'm not sure where the revision number comes from.


celskeggs:
Yeah. The changelog defines the current version of the debian package.

krawthekrow:
Changelog updated. It feels like I should have done it with a script instead of manually.

celskeggs:
The command you're looking for is probably [this one](https://manpages.debian.org/stretch/devscripts/debchange.1.en.html).

celskeggs:
For the sake of posterity and reproducibility, can you describe the process you used to verify this upstream code?

Also, this system works better if you sign your own commits. Do you have a PGP key? I can also sign it for you with my key, but then lots of things look slightly weird, such as this PR appearing as "closed" instead of "merged".

You will also probably want to rebase this commit against the new master branch.

krawthekrow:
Isn't that already covered in #106? Or do you mean stuff like using git or computing the SHA512 hash?

I don't have a PGP key, but I can change that.

andersk:
@krawthekrow The question is not how someone who already trusts our SHA512SUM.UPSTREAM can re-verify the hash you added. The question is how you verified that you put the right hash into SHA512SUM.UPSTREAM in the first place (e.g. which upstream GPG signatures you checked).

krawthekrow:
Source code downloaded from https://github.com/coreos/etcd/releases/tag/v3.2.9 (v3.2.9.tar.gz), which was signed with the CoreOS Application Signing Key (https://coreos.com/security/app-signing-key/) verified against trust.txt. The Github release was signed by Brandon Philips, but we don't have his signing key fingerprint in trust.txt.

krawthekrow:
I did `tar -xzvf ` and then `tar -cJf`, which was quite silly in retrospect. I've updated both repositories so that the procedure used is the same as the one you described, except with `gzip -d v0.9.0.tar.gz` instead of `ungzip 0.9.0.tar.gz`. For reference, I'm using gzip version 1.6 and xz version 5.1.0a.

celskeggs:
Still can't reproduce this. I'm using the same version of gzip and xz. (Also, woops, I mistyped 'gunzip' as 'ungzip', hence, the confusion.)

PR #113, opened 2017-10-10
Upgrade flannel to 0.9.0

krawthekrow:
Once again, I don't see a revision number anywhere. If I understand correctly, the revision number for flannel appears to be referenced in two places in the repository, and the current value for the revision number is different in both references.


celskeggs:
When you increment the version number, you should reset the revision number to 1. The revision is usually a counter of incremental changes to a downstream repackaging of upstream code, so we increment it when we make a local change and there's no new upstream release.

And yeah, the -2 in the flannel.yaml file is a mistake. It should be -3. Since you're changing the version, it makes sense for you to also change the revision.

krawthekrow:
Changes to revisions included.

celskeggs:
For the sake of posterity and reproducibility, can you describe the process you used to verify this upstream code?

Also, this system works better if you sign your own commits. Do you have a PGP key? I can also sign it for you with my key, but then lots of things look slightly weird, such as this PR appearing as "closed" instead of "merged".

You will also want to rebase this commit against the new master branch.

celskeggs:
I can't reproduce the hash that you provided.

The steps I used were the following:

    $ wget https://github.com/coreos/flannel/archive/v0.9.0.tar.gz
    $ ungzip v0.9.0.tar.gz
    $ xz v0.9.0.tar
    $ sha512sum v0.9.0.tar.xz

Can you provide more precise instructions for what you did to get this hash?

krawthekrow:
I did `tar -xzvf ` and then `tar -cJf`, which was quite silly in retrospect. I've updated both repositories so that the procedure used is the same as the one you described, except with `gzip -d v0.9.0.tar.gz` instead of `ungzip 0.9.0.tar.gz`. For reference, I'm using gzip version 1.6 and xz version 5.1.0a.

celskeggs:
Did you actually update the hash in the file? It still appears to be incorrect still -- and I think the same one as before.

celskeggs:
Can you rebase this?

krawthekrow:
Rebased.

celskeggs:
Please file a corresponding pull request on homeworld-upstream, making sure that it's rebased.

(Sorry for the delay -- I've been hosed.)

krawthekrow:
Oops, sorry, I keep having the impression that I had already filed that PR. It should be filed now.

krawthekrow:
Fixed. Sorry I didn't notice it earlier either.

celskeggs:
I forgot to change my review back to "approved", but it's merged anyway.

PR #115, opened 2017-10-10
homeworld-admin-tools: Extract iso without sudo, using libarchive-tools

andersk:
A normal build chroot doesn’t (and shouldn’t) give sudo access to the build process, so the build was failing here.


andersk:
Added `libarchive-tools` to `Depends` too, since this isn’t necessarily just run at build time.

celskeggs:
This does not work in jessie; libarchive-tools isn't available until stretch. Is there an easy way around this, or do we need to make stretch our new minimum version?

andersk:
The package is named `bsdtar` in jessie.

Issue #116, opened 2017-10-10
Audit Lintian warnings for Debian packages


andersk:
If you configure `sbuild` to run `lintian` in `~/.sbuildrc`:

```
$run_lintian = 1;
1;
```

then you get a bunch of messages about the built Debian packages, some of which we should pay attention to ([documentation](https://lintian.debian.org/)).

> W: homeworld-admin-tools source: debhelper-but-no-misc-depends homeworld-admin-tools
> W: homeworld-apt-setup source: debhelper-but-no-misc-depends homeworld-apt-setup
> W: homeworld-bootstrap-registry source: debhelper-but-no-misc-depends homeworld-bootstrap-registry
> W: homeworld-etcd source: debhelper-but-no-misc-depends homeworld-etcd
> W: homeworld-hyperkube source: debhelper-but-no-misc-depends homeworld-hyperkube
> W: homeworld-hyperkube: missing-depends-line
> W: homeworld-keysystem source: debhelper-but-no-misc-depends homeworld-keysystem
> E: homeworld-keysystem: missing-dependency-on-libc needed by usr/bin/keyclient and 5 others
> W: homeworld-knc source: debhelper-but-no-misc-depends homeworld-knc
> E: homeworld-knc: missing-dependency-on-libc needed by usr/bin/knc
> W: homeworld-rkt source: package-depends-on-hardcoded-libc homeworld-rkt depends
> W: homeworld-rkt source: debhelper-but-no-misc-depends homeworld-rkt
> W: homeworld-services source: debhelper-but-no-misc-depends homeworld-services

- [x] All packages need `${misc:Depends}`.
- [x] Packages with binaries or libraries need `${shlibs:Depends}`.

> W: homeworld-admin-tools source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-apt-setup source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-bootstrap-registry source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-etcd source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-hyperkube source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-keysystem source: ancient-standards-version 3.9.6 (current is 3.9.8)
> W: homeworld-knc source: ancient-standards-version 3.9.6 (current is 3.9.8)
[...]

PR #119, opened 2017-10-10
Some Debian packaging cleanup

andersk:
Addresses the easiest parts of #116.


celskeggs:
Might be worth signing that last commit.

PR #135, opened 2017-12-03
Add disaster key encryption and fix cluster deployment for dc6-proto1

celskeggs:
There are lots of changes here.

PR #140, opened 2017-12-03
Fix spire setup supervisor-ssh crashing when run twice

### LINKED ISSUE #131: spire setup supervisor-ssh is not idempotent
Removing the `authorized_keys` file twice fails the second time.

krawthekrow:
Check that the file exists before mv-ing it. Fixes #131.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/debian/changelog
+++ b/building/build-debs/homeworld-admin-tools/debian/changelog
@@ -1,3 +1,9 @@
+homeworld-admin-tools (0.1.22) stretch; urgency=medium
+
+  * Updated debian release
+
+ -- Mark Theng <mtheng@mit.edu>  Sun, 03 Dec 2017 18:43:38 -0500
+
 homeworld-admin-tools (0.1.21) stretch; urgency=medium
 
   * Updated debian release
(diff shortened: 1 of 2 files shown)


celskeggs:
Merging this is blocked on getting #135 merged.

celskeggs:
I accidentally published a different change as 0.1.21 -- I would renumber it, but it's not possible to rewrite a version in the repository once it's been published, so you'll need to renumber this to 0.1.22, and rebase it against the HEAD of `better-addon-deploy`.

Since there's a bit more review before #145 gets merged, and so the head of `better-addon-deploy` may change, you might want to wait until #145 has been merged, and then just rebase against master.

ichung:
Is there a better way to handle debian version numbers?  It seems like we will have a recurring problem with them every time we have concurrent changes.

celskeggs:
@ichung: I don't have a better solution. Any ideas?

celskeggs:
Can you rebase this and renumber the version?

celskeggs:
(Also, sometimes I don't get a notification when a commit is pushed, so when you've resolved things, it's usually fastest to assign this PR back to me, so that I can take a look quickly.)

krawthekrow:
Oh, that's a good idea. (Also, I should stop relying on Zephyr as a reliable communication platform.)

celskeggs:
Zephyr usually works fine, but has edge cases, especially when other conversations are happening.

PR #146, opened 2017-12-17
Add whois as spire dependency

krawthekrow:
Spire uses mkpasswd, which is part of the whois package.


celskeggs:
You'll need to rebase this against krawthekrow:fix-supervisor-ssh-not-idempotent and renumber the version. (You could just rebase against master, but then you'll have to rebase when your other PR gets merged, so no point.)

PR #154, opened 2017-12-17
Add instructions to pull upstream before build

rsthomp:
This fixes the symbolic links in the build-helper folder, which otherwise keep the helpers from building.


docs/build.md
```diff
@@ -9,6 +9,10 @@ Install packages:
 
 You will need to set up a sbuild chroot. (See https://github.com/sipb/homeworld/issues/80)
 
+# Pull required libraries and images from upstream
+    $ cd //building
+    $./pull-upstream.sh
```
celskeggs: This is missing a space between the `$` and the `.`.

celskeggs:
These need to be signed with your key instead of GitHub's key.

celskeggs:
Looks good, but I need to verify your GPG key in person.

building/build-helpers/helper-go/.gitignore
```diff
@@ -1 +1 @@
-go
+go*
```
andersk: This should be more specific so as not to match `go1.4-bootstrap-20170531.tar.xz` and `go1.8.3.src.tar.xz` which are already in the repository as symlinks:

```
/go*/
```

(Leading slash: only match immediate children of this directory `helper-go`. Trailing slash: only match those that are themselves directories.)

celskeggs: I'm addressing this in a separate commit for the sake of moving things along.

celskeggs: I'm addressing this in a separate commit for the sake of moving things along.

Issue #160, opened 2017-12-17
spire should make errors more obvious


ichung:
In red, say.

Issue #153, opened 2017-12-17
Vague error message for "spire seq keysystem"


krawthekrow:
At "verify that the keygateway is responsive", if Kerberos tickets have not been generated, spire fails with error message "keygateway check failed: Command '['keyreq', 'check', '/tmp/...', 'egg-sandwich.mit.edu:20557']' returned non-zero exit status 1". This should be more descriptive.

Issue #152, opened 2017-12-17
spire seq keysystem not idempotent


krawthekrow:
"confirm no existing keytab on egg-sandwich" fails when "spire seq keysystem" is run a second time. Maybe remove this check altogether?

PR #169, opened 2017-12-20
Improve seq keysystem

### LINKED ISSUE #151: spire seq keysystem verification step randomly crashing
"verify that keyserver static files can be fetched" sometimes fails with "ConnectionRefusedError: [Errno 111] Connection refused". Possibly a timeout problem.

krawthekrow:
Fixes #151, #152 and #153. ~For #152, I simply removed the line that checks if a keytab already exists -- is there a reason why it was there to begin with? If so, I could make it so that it checks for and ignores the error in seq.py instead. I can't quite think of a case when you wouldn't want to update the keytab if there's one already there.~ (Resolved in review)

### DIFF
--- a/building/build-debs/homeworld-admin-tools/debian/changelog
+++ b/building/build-debs/homeworld-admin-tools/debian/changelog
@@ -1,3 +1,9 @@
+homeworld-admin-tools (0.1.25) stretch; urgency=medium
+
+  * Updated debian release
+
+ -- Mark Theng <mtheng@mit.edu>  Thu, 28 Dec 2017 14:33:36 -0500
+
 homeworld-admin-tools (0.1.24) stretch; urgency=medium
 
   * Updated debian release
--- a/building/build-debs/homeworld-admin-tools/src/access.py
+++ b/building/build-debs/homeworld-admin-tools/src/access.py
@@ -41,7 +41,7 @@ def create_or_rotate_custom_ssh_key(interval=DEFAULT_ROTATE_INTERVAL, bits=DEFAU
     return keypath
 
 
-def call_keyreq(command, *params, collect=False):
+def call_keyreq(keyreq_command, *params, collect=False):
     config = configuration.get_config()
     keyserver_domain = config.keyserver.hostname + "." + config.external_domain + ":20557"
 
@@ -50,7 +50,15 @@ def call_keyreq(command, *params, collect=False):
     with tempfile.TemporaryDirectory() as tdir:
         https_cert_path = os.path.join(tdir, "server.pem")
         util.writefile(https_cert_path, authority.get_pubkey_by_filename("./server.pem"))
-        return invoke_variant(["keyreq", command, https_cert_path, keyserver_domain] + list(params))
+        try:
+            return invoke_variant(["keyreq", keyreq_command, https_cert_path, keyserver_domain] + list(params))
+        except subprocess.CalledProcessError as e:
+            if e.returncode == 1:
+                fail_hint = "do you have valid kerberos tickets?\n" \
+                    "or, your connection to the server might be faulty.\n" \
+                    "or, the server's keygateway might be broken."
+                command.fail("keyreq failed: %s" % e, fail_hint)
+            raise e
 
 
 def renew_ssh_cert() -> str:
@@ -67,10 +75,9 @@ def access_ssh(add_to_agent=False):
     if add_to_agent:
         # TODO: clear old identities
         if subprocess.call(["ssh-add", "--", keypath]) != 0:
-            print("ssh-add returned non-zero exit code. do you have a ssh-agent?")
-            print("or, perhaps your agent is broken; consider killing it and launching a new one.")
-            command.fail("*** ssh-add failed! ***")
-
+            fail_hint = "ssh-add returned non-zero exit code. do you have a ssh-agent?\n" \
+                "or, perhaps your agent is broken; consider killing it and launching a new one."
+            command.fail("*** ssh-add failed! ***", fail_hint)
 
 def access_ssh_with_add():
     access_ssh(add_to_agent=True)
--- a/building/build-debs/homeworld-admin-tools/src/command.py
+++ b/building/build-debs/homeworld-admin-tools/src/command.py
@@ -1,12 +1,19 @@
 import inspect
 
 
+ANSI_ESCAPE_CODE_RED = "\x1b[1;31m"
+ANSI_ESCAPE_CODE_YELLOW = "\x1b[1;33m"
+ANSI_ESCAPE_CODE_RESET = "\x1b[1;0m"
+
+
 class CommandFailedException(Exception):
-    pass
+    def __init__(self, message, hint):
+        super().__init__(message)
+        self.hint = hint
 
 
-def fail(message: str) -> None:
-    raise CommandFailedException(str(message))
+def fail(message: str, hint: str = None) -> None:
+    raise CommandFailedException(message, hint)
 
 
 def mux_map(desc: str, mapping: dict):
@@ -76,5 +83,7 @@ def main_invoke(command, params):
         invoke(params)
         return 0
     except CommandFailedException as e:
-        print('\x1b[1;31m' + 'command failed: ' + str(e) + '\x1b[0m')
+        print(ANSI_ESCAPE_CODE_RED + 'command failed: ' + str(e) + ANSI_ESCAPE_CODE_RESET)
+        if e.hint is not None:
+            print(ANSI_ESCAPE_CODE_YELLOW + e.hint + ANSI_ESCAPE_CODE_RESET)
         return 1
--- a/building/build-debs/homeworld-admin-tools/src/setup.py
+++ b/building/build-debs/homeworld-admin-tools/src/setup.py
@@ -1,6 +1,6 @@
 import os
-import tempfile
 import time
+import subprocess
 
 import authority
 import command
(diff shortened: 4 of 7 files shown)


building/build-debs/homeworld-admin-tools/src/setup.py
```diff
@@ -129,7 +129,6 @@ def setup_keygateway(ops: Operations, config: configuration.Config) -> None:
         # keytab is stored encrypted in the configuration folder
         keytab = os.path.join(configuration.get_project(), "keytab.%s.crypt" % node.hostname)
         decrypted = keycrypt.gpg_decrypt_to_memory(keytab)
-        ops.ssh("confirm no existing keytab on @HOST", node, "test", "!", "-e", KEYTAB_PATH)
```
celskeggs: The original point here was to avoid overwriting a newer keytab with an older keytab.

We have a few options:

 * Your current change, which allows for the possibility of overwriting and losing a keytab.
 * Silently refuse to overwrite the keytab, which will cause pains when the keytab is updated, but will allow for re-deploying the same cluster.
 * Refuse to overwrite the keytab only if it has changed, which will cause pain (but not as much) when the keytab is updated, and allow for re-deploying the same cluster, but will be more complicated.
 * Refuse to overwrite the keytab only if it has changed and the updated keytab would have an older kvno than the last keytab, which will be the best solution, but will be much more complicated.

I would recommend the third solution, but I'm interested in any thoughts you have.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -75,7 +75,7 @@ def check_keygateway():
     try:
         access.call_keyreq("check")
     except subprocess.CalledProcessError as e:
-        command.fail("keygateway check failed: %s" % e)
+        command.fail("keygateway check failed (do you have valid Kerberos tickets?): %s" % e)
```
celskeggs: It might be worth detecting the exact failure mode ('returned non-zero exit status 1') and only providing the extra information in that case, and then also being louder about the the error, rather than sticking it in the middle of a longer error message.

Maybe an error like:

    keygateway check failed: [...]
    ** do you have valid kerberos tickets? **

celskeggs:
Sorry for not getting back to you sooner on this.

building/build-debs/homeworld-admin-tools/src/setup.py
krawthekrow: Would it be better to merge the keytabs? From what I understand, you would mainly be concerned about overwriting the existing keytab if there already exists a different keytab for other reasons. Maybe Spire can make a backup of the existing keytab and spit out a warning at the end of the session.

Or is the concern that the administrator may accidentally run the command with an outdated keytab? It doesn't sound like that would be a problem if the latest keytab is stored in, for example, Git.

ichung: > ...the administrator may accidentally run the command with an outdated keytab? It doesn't sound like that would be a problem...

That sounds like precisely the sort of thing people will do accidentally.

krawthekrow: I meant that if he does do that, he can just get the correct keytab from Git and replace it.

celskeggs: Only if they realize it before their certificates expire. Otherwise, no one can regain access to the cluster until someone goes and drags the disaster recovery keys out of cold storage.

krawthekrow: Right, it's encrypted with the disaster recovery key. Would merging the keytabs work here, then? It should be relatively simple to do it with ktutil; it's possible to scrape the kvnos from klist, but that feels ugly.

krawthekrow:
"Fixed" the other issue mentioned in the review; now fixes #160 too.

building/build-debs/homeworld-admin-tools/src/setup.py
celskeggs: Being encrypted with the disaster recovery key has nothing to do with it -- you'd need to pull the DR key out of storage simply because you have no other way to access the cluster.

Merging the keytabs might work, but I don't want that to silently occur. I think your idea of the third solution plus a force-overwrite option sounds good to me.

building/build-debs/homeworld-admin-tools/src/command.py
```diff
@@ -5,8 +5,11 @@ class CommandFailedException(Exception):
     pass
 
 
-def fail(message: str) -> None:
-    raise CommandFailedException(str(message))
+def fail(message: str, hint: str = None) -> None:
+    formatted_message = '\033[91m' + str(message)
+    if hint is not None:
+        formatted_message += '\n\033[93m' + str(hint)
+    raise CommandFailedException(formatted_message)
```
celskeggs: Besides the fact that this is being fixed elsewhere (note that #160 is claimed), it probably doesn't make sense to add color at this point: it probably makes more sense to add color when CommandFailedException is caught on the other end.

I do like your idea of having a "hint" section, however, so I recommend you expand CommandFailedException to have that as an additional attribute and print it out when CommandFailedException is caught.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -75,7 +75,9 @@ def check_keygateway():
     try:
         access.call_keyreq("check")
     except subprocess.CalledProcessError as e:
-        command.fail("keygateway check failed: %s" % e)
+        command.fail("keygateway check failed: %s" % e,
+            '** do you have valid kerberos tickets? **' if
+            e.returncode == 1 else None)
```
celskeggs: I would avoid using an inline `if` here, for the sake of readability, but it doesn't matter that much.

krawthekrow:
Alright, everything has been responded to. I'll update it to use check_ssh after #171 gets finalized.

building/build-debs/homeworld-admin-tools/src/setup.py
```diff
@@ -122,18 +122,37 @@ def admit_keyserver(ops: Operations, config: configuration.Config) -> None:
...
+                    subprocess.check_call(["ssh", "root@%s.%s" % (node.hostname, config.external_domain), "test", "!", "-e", KEYTAB_PATH])
+                except subprocess.CalledProcessError as e_test:
+                    if e_test.returncode != 1:
+                        command.fail(e_test)
+                    print("existing keytab found")
+                    existing_keytab = subprocess.check_output(["ssh", "root@%s.%s" % (node.hostname, config.external_domain), "cat", KEYTAB_PATH])
+                    if existing_keytab != decrypted:
+                        command.fail("existing keytab does not match local keytab")
+                    print("existing keytab matches")
+                    return
```
celskeggs: You can probably encode this as a single ssh command, rather than needing to round-trip it multiple times and have all of this helper code involved.

building/build-debs/homeworld-admin-tools/src/setup.py
```diff
@@ -212,6 +231,7 @@ def wrap_param_tx(params):
     "keyserver": wrapop("deploy keys and configuration for keyserver; start keyserver", setup_keyserver),
     "self-admit": wrapop("admit the keyserver into the cluster during bootstrapping", admit_keyserver),
     "keygateway": wrapop("deploy keytab and start keygateway", setup_keygateway),
+    "update-keytab": wrapop("update keytab and restart keygateway", update_keytab),
```
celskeggs: These two commands should have clearer names. Also, keep in mind that the parameter to setup_keygateway will be treated as an optional command-line parameter, as far as I remember, which might not be what you want.

building/build-debs/homeworld-admin-tools/src/setup.py
krawthekrow: Good idea, will implement.

building/build-debs/homeworld-admin-tools/src/setup.py
krawthekrow: What names do you suggest? "Set up keygateway" and "update the keytab" sounds like exactly what those commands are for. Would it be better to actually have force-overwrite as an optional string parameter, so that the administrator would call "spire setup keygateway --force-overwrite" to update the keytab?

Also thanks for noting that, will fix.

celskeggs: I'd name it `update-keygateway`, so that it's clear that the operations are very similar.

krawthekrow:
Not a trivial rebase. Moved upload_bytes to ssh.py, added type hints to functions in ssh.py, made hint yellow, and updated access_ssh to use hint system.

building/build-debs/homeworld-admin-tools/src/access.py
```diff
@@ -67,10 +67,8 @@ def access_ssh(add_to_agent=False):
     if add_to_agent:
         # TODO: clear old identities
         if subprocess.call(["ssh-add", "--", keypath]) != 0:
-            print("ssh-add returned non-zero exit code. do you have a ssh-agent?")
-            print("or, perhaps your agent is broken; consider killing it and launching a new one.")
-            command.fail("*** ssh-add failed! ***")
-
+            fail_hint = "ssh-add returned non-zero exit code. do you have a ssh-agent?\nor, perhaps your agent is broken; consider killing it and launching a new one."
```
celskeggs: Might want to have this be multiple lines for readability, either with `"""..."""` notation or by having a second line with `fail_hint +=`.

building/build-debs/homeworld-admin-tools/src/command.py
```diff
@@ -76,5 +78,9 @@ def main_invoke(command, params):
         invoke(params)
         return 0
     except CommandFailedException as e:
+        # red
         print('\x1b[1;31m' + 'command failed: ' + str(e) + '\x1b[0m')
+        if e.hint is not None:
+            # yellow
+            print('\x1b[1;33m' + e.hint + '\x1b[0m')
```
celskeggs: We could make this clearer by putting our magic strings into named globals.

building/build-debs/homeworld-admin-tools/debian/changelog
```diff
@@ -1,3 +1,9 @@
+homeworld-admin-tools (0.1.25); urgency=medium
```
celskeggs: This is missing the `stretch` keyword.

krawthekrow: Oops, thanks!

krawthekrow:
Addressed issues. I also moved the error checking for check_keygateway into call_keyreq, since the same problems occur everywhere that call_keyreq is called, and those places do not appear to handle keyreq errors.

Issue #141, opened 2017-12-04
spire should have a version command


celskeggs:
This should provide the commit hash that spire was built from, along with the debian package version.

Preferably, it would also append `-dirty` or something to the commit hash if any changes have been made to the working tree.

ed1d1a8d:
Would be interested in working on.

celskeggs:
Fixed by #170.

PR #170, opened 2017-12-27
Add spire version

ed1d1a8d:
Addresses https://github.com/sipb/homeworld/issues/141.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/build-package.sh
+++ b/building/build-debs/homeworld-admin-tools/build-package.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
-set -eu
+set -e -u
 source ../common/package-build-helpers.sh
+
+function cleanup {
+    rm -f resources/GIT_VERSION
+}
+trap cleanup EXIT
+
+git rev-parse HEAD | tr -d '\n' >resources/GIT_VERSION
+[[ -n "$(git status --porcelain)" ]] && echo -n "-dirty" >>resources/GIT_VERSION
+echo >>resources/GIT_VERSION
 
 upstream "debian-9.2.0-amd64-mini.iso"
 build
--- a/building/build-debs/homeworld-admin-tools/debian/rules
+++ b/building/build-debs/homeworld-admin-tools/debian/rules
@@ -21,7 +21,7 @@ override_dh_auto_build:
 
 # main packaging script based on dh7 syntax
 %:
-	dh $@ 
+	dh $@
 
 # debmake generated override targets
 # This is example for Cmake (See http://bugs.debian.org/641051 )
--- a/building/build-debs/homeworld-admin-tools/inner-build.sh
+++ b/building/build-debs/homeworld-admin-tools/inner-build.sh
@@ -3,6 +3,21 @@ set -e -u
 
 rm -rf spire spire.zip src/resources src/__pycache__ resources/__pycache__
 cp -RT resources src/resources
+
+if [[ -v VERSION ]]
+then
+    echo "$VERSION" >src/resources/DEB_VERSION
+else
+    echo "not a debian build" >src/resources/DEB_VERSION
+fi
+
+if [[ ! -e src/resources/GIT_VERSION ]]
+then
+    git rev-parse HEAD | tr -d '\n' >src/resources/GIT_VERSION
+    [[ -n "$(git status --porcelain)" ]] && echo -n "-dirty" >>src/resources/GIT_VERSION
+    echo >>src/resources/GIT_VERSION
+fi
+
 (cd src && zip -r ../spire.zip *)
 
 python3 spire.zip iso regen-cdpack debian-9.2.0-amd64-mini.iso src/resources/debian-9.2.0-cdpack.tgz
--- a/building/build-debs/homeworld-admin-tools/resources/__init__.py
+++ b/building/build-debs/homeworld-admin-tools/resources/__init__.py
@@ -2,9 +2,11 @@ import pkgutil
 
 
 def get_resource(name: str) -> bytes:
-    b = pkgutil.get_data(__package__, name)
+    try:
+        b = pkgutil.get_data(__package__, name)
+    except OSError:
+        raise Exception("no such embedded resource: %s" % name) from None
     if b is None:
-        raise Exception("no such embedded resource: %s" % name)
+        raise Exception("package cannot be located or loaded: %s" % __package__)
     assert type(b) == bytes  # todo: remove this
     return b
-
--- a/building/build-debs/homeworld-admin-tools/src/__main__.py
+++ b/building/build-debs/homeworld-admin-tools/src/__main__.py
@@ -12,6 +12,7 @@ import infra
 import keys
 import seq
 import deploy
+import version
 
 main_command = command.mux_map("invoke a top-level command", {
     "iso": iso.main_command,
@@ -28,6 +29,7 @@ main_command = command.mux_map("invoke a top-level command", {
     "infra": infra.main_command,
     "seq": seq.main_command,
     "deploy": deploy.main_command,
+    "version": version.main_command
 })
 
 
--- a/building/build-debs/homeworld-admin-tools/src/version.py
+++ b/building/build-debs/homeworld-admin-tools/src/version.py
@@ -1,0 +1,14 @@
+import os
+import command
+from resources import get_resource
+
+
+def display_version():
+    deb_version = get_resource("DEB_VERSION").decode().rstrip()
+    print("Debian package version:", deb_version)
+
+    git_version = get_resource("GIT_VERSION").decode().rstrip()
+    print("Git commit hash:", git_version)
+
+
+main_command = command.wrap("display version info", display_version)
--- a/building/build-debs/homeworld-admin-tools/.gitignore
+++ b/building/build-debs/homeworld-admin-tools/.gitignore
@@ -2,3 +2,5 @@
 /spire
 debian-*-amd64-mini.iso
 debian-*-cdpack.tgz
+resources/GIT_VERSION
+src/resources/


ed1d1a8d:
The git version is only displayed if homeworld-admin-tools/build-package.sh is called. Calling inner-build does not yield the git version.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -1,6 +1,28 @@
 #!/bin/bash
-set -eu
+set -e -u
+rm -f resources/GIT_VERSION
```
celskeggs: There's no need to remove this file explicitly; ` >resources/GIT_VERSION` will overwrite it.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -1,6 +1,28 @@
...
+set +e
+git diff-index --quiet HEAD --
+GIT_DIFF_RETVAL=$?
+set -e
+
+GIT_ROOT="$(git rev-parse --show-toplevel)"
+set +e
+u="$(git ls-files --exclude-standard --others "${GIT_ROOT}")" && test -z "$u"
+GIT_LS_RETVAL=$?
+set -e
```
celskeggs: This seems rather complicated. Maybe something from https://unix.stackexchange.com/a/155077 might be easier?

building/build-debs/homeworld-admin-tools/debian/rules
```diff
@@ -17,11 +17,11 @@ include /usr/share/dpkg/default.mk
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
 override_dh_auto_build:
-	VERSION="${DEB_VERSION_UPSTREAM}" ./inner-build.sh
+	DEB_VERSION="${DEB_VERSION_UPSTREAM}" ./inner-build.sh
```
celskeggs: Is it necessary to rename this? It breaks consistency with the other debian/rules files.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -1,6 +1,28 @@
...
+then
+    echo -n "-dirty" >>resources/GIT_VERSION
+fi
+echo >>resources/GIT_VERSION
+
 source ../common/package-build-helpers.sh
 
 upstream "debian-9.2.0-amd64-mini.iso"
 build
+rm resources/GIT_VERSION
```
celskeggs: Might be worth doing this in a trap handler, so that it happens even if the build fails.

building/build-debs/homeworld-admin-tools/src/__main__.py
```diff
@@ -28,6 +29,7 @@
     "infra": infra.main_command,
     "seq": seq.main_command,
     "deploy": deploy.main_command,
+    "version": command.wrap("display version info", version.display_version)
```
celskeggs: For the sake of consistency and maintaining abstraction boundaries, it makes more sense to have `version` export a `main_command` attribute that equals `command.wrap("display version info", display_version)` instead of this.

building/build-debs/homeworld-admin-tools/src/version.py
```diff
@@ -0,0 +1,10 @@
+import os
+from resources import get_resource
+
+
+def display_version():
+    deb_version = get_resource("DEB_VERSION").decode("utf-8").rstrip()
+    print("Debian package version:", deb_version)
+
+    git_version = get_resource("GIT_VERSION").decode("utf-8").rstrip()
+    print("Build commit hash:", git_version)
```
celskeggs: We generally don't bother with specifying the codec when we call `.decode()`. Do we have a particular new reason to do that?

Also, does this correctly handle the case where we build directly with inner_build.sh and resources/GIT_VERSION is missing?

building/build-debs/homeworld-admin-tools/build-package.sh
ed1d1a8d: We either need to remove it or gitignore `resources/GIT_VERSION` since we are checking if the working directory is clean immediately after. Which is preferable?

building/build-debs/homeworld-admin-tools/debian/rules
ed1d1a8d: Not necessary. It just makes the variable naming in inner-build.sh nicer. Do you want me to change it back?

building/build-debs/homeworld-admin-tools/src/__main__.py
ed1d1a8d: Will change.

building/build-debs/homeworld-admin-tools/build-package.sh
ed1d1a8d: WiIl simplify.

building/build-debs/homeworld-admin-tools/src/version.py
ed1d1a8d: Regarding your first point: I did not realize decode defaulted to utf-8. Will change the code to not specify codec.

ed1d1a8d: Regarding your second point: the code crashes when building only with inner-build.sh. What should be the behavior when just building with inner-build? Should it behave the same as outer-build?

ed1d1a8d:
What should be the behavior when just building with inner-build? Should it behave the same as build-package?

building/build-debs/homeworld-admin-tools/build-package.sh
celskeggs: gitignoring it seems like the right solution. We wouldn't want to check it in, anyway.

building/build-debs/homeworld-admin-tools/debian/rules
celskeggs: Yeah -- if you want to submit a separate change to improve the naming, that'd be good, but perhaps not in this PR.

building/build-debs/homeworld-admin-tools/src/version.py
celskeggs: Well, it should *definitely* not crash! Especially not delayed until runtime like that. A long-distance crash caused by building a different way is an unfortunate bug.

We also do call ./inner-build.sh directly during testing, so we should definitely support this condition. We should probably populate the resource in all cases.

With regards to the debian version, it probably makes more sense to only populate it when using ./build_package.sh, and then say "this was not built as part of a debian package" if ./inner_build.sh was used, either by detecting the lack of the resource or by sticking that string into the file.

It's probably worth, in any case, making sure that there's a clear error message if either of these is missing. It might be clear enough to just say "GIT_VERSION cannot be found" or something, which might already be the error message, but if the error message is more confusing than that, it might be worth adding extra code for a better message.

celskeggs:
Well, not *exactly* the same, but it should definitely succeed and build a working binary. See more specific comments on the review itself.

Once you've addressed the comments, you should assign this PR back to me so that I know I need to take care of it.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -15,4 +19,3 @@ source ../common/package-build-helpers.sh
```
celskeggs: These changes should probably be stuck below sourcing of package-build-helpers.sh.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -2,23 +2,13 @@
...
-git diff-index --quiet HEAD --
-GIT_DIFF_RETVAL=$?
-set -e
-
-GIT_ROOT="$(git rev-parse --show-toplevel)"
-set +e
-u="$(git ls-files --exclude-standard --others "${GIT_ROOT}")" && test -z "$u"
-GIT_LS_RETVAL=$?
+u="$(git status --porcelain)" && test -z "$u"
+GIT_DIRTY=$?
```
celskeggs: You realize you can just set `u` here and then use `test -z "$u"` down below (or its negation, perhaps) instead of comparing the retval, right?

ed1d1a8d:
I don't think I can edit the assignees list.

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -1,5 +1,20 @@
...
+trap cleanup EXIT
+cleanup
+
+set +e
+u="$(git status --porcelain)" && test -z "$u"
+GIT_DIRTY=$?
+set -e
+
+git rev-parse HEAD | tr -d '\n' >resources/GIT_VERSION
+[[ $GIT_DIRTY -ne 0 ]] && echo -n "-dirty" >>resources/GIT_VERSION
```
celskeggs: You realize you can just set u above and then use test -z "$u" down here (or its negation, perhaps) instead of comparing the retval, right?

building/build-debs/homeworld-admin-tools/build-package.sh
ed1d1a8d: Just wondering, why is this the best practice?

building/build-debs/homeworld-admin-tools/build-package.sh
ed1d1a8d: Yes, that sounds like a better idea.

building/build-debs/homeworld-admin-tools/build-package.sh
celskeggs: Good question. It's because source is kind of like import, so you do it at the top of the file for the same reason you usually do your imports at the top of a python file: the set of imports informs many things about what the file does, so by putting them at the top, the outline of what the file is doing becomes clearer to the reader.

celskeggs:
You should have access to edit the assignees list now.

ed1d1a8d:
I can confirm that I have access. Thanks.

building/build-debs/homeworld-admin-tools/resources/__init__.py
```diff
@@ -2,9 +2,11 @@
 
 
 def get_resource(name: str) -> bytes:
-    b = pkgutil.get_data(__package__, name)
+    try:
```
ed1d1a8d: Catches error when file not found. Per https://docs.python.org/3.5/library/pkgutil.html.
Should this go in a separate commit, since it is not directly related to spire version?

ed1d1a8d:
Works for both build-package and inner-build.

building/build-debs/homeworld-admin-tools/inner-build.sh
```diff
@@ -3,6 +3,21 @@ set -e -u
 
 rm -rf spire spire.zip src/resources src/__pycache__ resources/__pycache__
 cp -RT resources src/resources
+
+if [[ -v VERSION ]]
+then
+    echo "$VERSION" >src/resources/DEB_VERSION
+else
+    echo "Spire was not built as part of a Debian package." >src/resources/DEB_VERSION
```
celskeggs: Probably worth having a less verbose string would be good?

This is long:

    Debian package version: Spire was not built as part of a Debian package.

Compared to something like this:

    Debian package version: not a debian build

building/build-debs/homeworld-admin-tools/resources/__init__.py
celskeggs: Because it's a change involved in implementing `spire version`, it makes sense to include it in the same PR. You can choose whether it's cleaner in a separate commit within the PR -- whatever seems right.

building/build-debs/homeworld-admin-tools/inner-build.sh
```diff
@@ -3,6 +3,21 @@ set -e -u
...
+fi
+
+if [[ ! -e src/resources/GIT_VERSION ]]
+then
+    git rev-parse HEAD | tr -d '\n' >src/resources/GIT_VERSION
+    [[ -n "$(git status --porcelain)" ]] && echo -n "-dirty" >>src/resources/GIT_VERSION
+    echo >>src/resources/GIT_VERSION
+fi
```
celskeggs: I assume that the reason that this code is duplicated is because the git repository can't be accessed from within the build chroot?

In any case, you should make sure that inner_build.sh cleans up GIT_VERSION.

building/build-debs/homeworld-admin-tools/resources/__init__.py
celskeggs: Because it's a change involved in implementing spire version, it makes sense to include it in the same PR. You can choose whether it's cleaner in a separate commit within the PR -- whatever seems right.

building/build-debs/homeworld-admin-tools/inner-build.sh
ed1d1a8d: Will shorten to what you have there.

building/build-debs/homeworld-admin-tools/inner-build.sh
ed1d1a8d: I am writing it to src/resources/GIT_VERSION, which should already be cleaned up by the existing code. Though this is not done with a trap. Should I change it to use a trap?

building/build-debs/homeworld-admin-tools/resources/__init__.py
ed1d1a8d: Moved to separate commit.

building/build-debs/homeworld-admin-tools/inner-build.sh
celskeggs: Woops, you're right. It's good as it is.

celskeggs:
Actually, this needs to be rebased. Sorry 'bout that.

ed1d1a8d:
Rebased.

Issue #147, opened 2017-12-17
Vague error message for "spire seq ssh," should be more specific


rsthomp:
Error message for "spire seq ssh" :

    -- verify ssh access to supervisor  (5/5) --
    Traceback (most recent call last):
    File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
        "__main__", mod_spec)
    File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
        exec(code, run_globals)
    File "/usr/bin/spire/__main__.py", line 35, in <module>
    File "/usr/bin/spire/command.py", line 76, in main_invoke
    File "/usr/bin/spire/command.py", line 27, in invoke
    File "/usr/bin/spire/command.py", line 27, in invoke
    File "/usr/bin/spire/command.py", line 27, in invoke
    File "/usr/bin/spire/command.py", line 68, in invoke
    File "/usr/bin/spire/setup.py", line 35, in run_operations
    File "/usr/bin/spire/verify.py", line 90, in check_ssh_with_certs
    KeyError: 'SSH_AGENT_PID'

celskeggs:
I've edited this issue to format the traceback properly.

PR #171, opened 2017-12-27
A bunch of small improvements to deployment

celskeggs:
Fully tested.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/command.py
+++ b/building/build-debs/homeworld-admin-tools/src/command.py
@@ -10,25 +10,25 @@ def fail(message: str) -> None:
 
 
 def mux_map(desc: str, mapping: dict):
-    def usage(err: str = "no command") -> None:
+    def usage() -> None:
         print("commands:")
         for name, (desc, subinvoke) in sorted(mapping.items()):
             print("  %s: %s" % (name, desc))
-        if err is not None:
-            fail(err)
 
     def invoke(params):
         if not params:
             usage()
+            print("no command")
         elif params[0] not in mapping:
-            usage("unknown command: %s" % params[0])
+            usage()
+            fail("unknown command: %s" % params[0])
         else:
             desc, subinvoke = mapping[params[0]]
             subinvoke(params[1:])
 
     if "usage" not in mapping:
         mapping = dict(mapping)
-        mapping["usage"] = ("ask for this usage info", lambda _: usage(None))
+        mapping["usage"] = ("ask for this usage info", lambda _: usage())
 
     return desc, invoke
 
@@ -76,5 +76,5 @@ def main_invoke(command, params):
         invoke(params)
         return 0
     except CommandFailedException as e:
-        print(e)
+        print('\x1b[1;31m' + 'command failed: ' + str(e) + '\x1b[0m')
         return 1
--- a/building/build-debs/homeworld-admin-tools/src/iso.py
+++ b/building/build-debs/homeworld-admin-tools/src/iso.py
@@ -29,11 +29,11 @@ def regen_cdpack(source_iso, dest_cdpack):
         subprocess.check_call(["tar", "-czf", dest_cdpack, "-C", d, os.path.basename(cddir)])
 
 
-def add_password_to_log(password):
+def add_password_to_log(password, creation_time):
     passwords = os.path.join(configuration.get_project(), "passwords")
     if not os.path.isdir(passwords):
         os.mkdir(passwords)
-    passfile = os.path.join(passwords, "at-%s.gpg" % datetime.datetime.now().isoformat())
+    passfile = os.path.join(passwords, "at-%s.gpg" % creation_time)
     util.writefile(passfile, keycrypt.gpg_encrypt_in_memory(password))
 
 
@@ -68,9 +68,10 @@ def gen_iso(iso_image, authorized_key, cdpack=None):
 
         preseeded = resource.get_resource("preseed.cfg.in")
         generated_password = util.pwgen(20)
-        add_password_to_log(generated_password)
+        creation_time = datetime.datetime.now().isoformat()
+        add_password_to_log(generated_password, creation_time)
         print("generated password added to log")
-        preseeded = preseeded.replace(b"{{HASH}}", util.mkpasswd(generated_password))
+        preseeded = preseeded.replace(b"{{HASH}}", util.mkpasswd(generated_password)).replace(b"{{BUILDDATE}}", creation_time.encode())
         util.writefile(os.path.join(d, "preseed.cfg"), preseeded)
 
         inclusion += ["sshd_config.new", "preseed.cfg"]
--- a/building/build-debs/homeworld-admin-tools/src/verify.py
+++ b/building/build-debs/homeworld-admin-tools/src/verify.py
@@ -44,6 +44,16 @@ def check_keystatics():
     print("pass: keyserver serving correct static files")
 
 
+def check_ssh(node, command):
+    config = configuration.Config.load_from_project()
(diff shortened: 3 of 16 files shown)


building/build-debs/homeworld-admin-tools/src/command.py
```diff
@@ -10,18 +10,21 @@ def fail(message: str) -> None:
 
 
 def mux_map(desc: str, mapping: dict):
-    def usage(err: str = "no command") -> None:
+    def usage(err: str = "no command", loud_err: bool = False) -> None:
```
krawthekrow: Seems like a rather convoluted way to achieve something simple. Maybe take error handling out of usage (so that the function's only job is to print the command help sheet)? The two of only three times when usage is called with an error message have different error messages anyway.

building/build-debs/homeworld-admin-tools/src/iso.py
```diff
@@ -29,11 +29,11 @@ def regen_cdpack(source_iso, dest_cdpack):
         subprocess.check_call(["tar", "-czf", dest_cdpack, "-C", d, os.path.basename(cddir)])
 
 
-def add_password_to_log(password):
+def add_password_to_log(password, now):
```
krawthekrow: Minor, but "creation_time" would be a better variable name here.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -90,7 +98,7 @@ def check_ssh_with_certs(hostname=None):
     del env["SSH_AGENT_PID"]
     keypath = access.renew_ssh_cert()
     try:
-        result = subprocess.check_output(["ssh", "-i", keypath, "root@%s.%s" % (hostname, config.external_domain), "echo confirmed"], env=env)
+        result = subprocess.check_output(["ssh", "-o", "StrictHostKeyChecking=yes", "-i", keypath, "root@%s.%s" % (hostname, config.external_domain), "echo confirmed"], env=env)
```
krawthekrow: The SSH parameters are repeated a lot everywhere in special cases like this one; maybe it would be good to make them a global variable.

building/build-debs/homeworld-admin-tools/src/command.py
```diff
@@ -76,5 +79,5 @@ def main_invoke(command, params):
         invoke(params)
         return 0
     except CommandFailedException as e:
-        print(e)
+        print('\x1b[1;31m' + str(e) + '\x1b[0m')
```
krawthekrow: Could we add a line saying "Command failed" or something? Currently it is not clear that it is an error, and I feel that this clarity should not solely depend on color.

building/build-debs/homeworld-admin-tools/src/command.py
celskeggs: That is... a very good idea and I'm not sure why I didn't do that.

building/build-debs/homeworld-admin-tools/src/iso.py
celskeggs: Will fix.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -44,6 +44,16 @@ def check_keystatics():
     print("pass: keyserver serving correct static files")
 
 
+def check_ssh(node, command):
+    config = configuration.Config.load_from_project()
```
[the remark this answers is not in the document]
celskeggs: A session object wouldn't be worthwhile, because this isn't a long-running process. It's probably worth separating out a helper like `configuration.get_config()`, which could cache the configuration -- but I don't think it matters much, because it's not like loading the configuration is a bottleneck.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -44,6 +44,16 @@ def check_keystatics():
...
+    subprocess.check_call(["ssh", "-o", "StrictHostKeyChecking=yes", "root@%s.%s" % (node.hostname, config.external_domain), "--"] + command)
```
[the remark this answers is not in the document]
celskeggs: Agreed on both counts.

building/build-debs/homeworld-admin-tools/src/command.py
celskeggs: I did that, but it seemed overdone in a number of cases, so I removed it. Do you think it's worthwhile anyway?

I mostly got things like:

    $ spire
    commands.
        [...]
    no command
    * * * * COMMAND FAILED * * * *

Though it's possible that there's a better solution.

celskeggs:
I'll try to take care of this tomorrow.

building/build-debs/homeworld-admin-tools/src/verify.py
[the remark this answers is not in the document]
krawthekrow: It's not so much performance as the fact that you're effectively using the config file as a global variable. This is not consistent with, for example, setup.py where functions take in the configuration as a parameter -- if I were to call such a function with a modified configuration, I would expect it to behave according to that configuration. If you're intent on configuration being global, it would be more explicit to stick it in a global variable and reference that everywhere the configuration is used.

By passing in a session object, I meant something like wrapop in setup.py. In fact, Operations could itself serve as the session object, with some changes.

Alternatively, the configuration could just be passed in as an additional parameter to check_ssh.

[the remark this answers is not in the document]
celskeggs: The reason that the functions have that in setup.py is just for convenience -- all of those methods were originally intended to *only* be called automatically. If that's confusing, the correct solution is to eliminate those parameters, and just letting everyone use get_config().

Configuration being global is intentional, because it avoids passing around unnecessary state and simplifies the code.

I'll go and eliminate some of the places where it's stored or passed around, to make this more clear.

building/build-debs/homeworld-admin-tools/src/command.py
krawthekrow: It would be overdone in a number of cases, but there are also a number of cases where it isn't, like "no keyserver found" or "expected kubelet to be ready". It should be sufficient to prepend "command failed: " in front of the error message, to make it clear that it's an error message.

celskeggs:
I think that covers everything. I haven't re-tested this, in case you have more feedback, but I'll grab the cluster lock and test another deploy once any remaining feedback is resolved.

celskeggs:
I added a trivial fix for #147, because it was only adding two lines.

building/build-debs/homeworld-admin-tools/src/verify.py
```diff
@@ -80,17 +79,19 @@ def check_keygateway():
...
-    del env["SSH_AUTH_SOCK"]
-    del env["SSH_AGENT_PID"]
+    if "SSH_AUTH_SOCK" in env:
+        del env["SSH_AUTH_SOCK"]
+    if "SSH_AGENT_PID" in env:
+        del env["SSH_AGENT_PID"]
     keypath = access.renew_ssh_cert()
     try:
-        result = subprocess.check_output(["ssh", "-i", keypath, "root@%s.%s" % (hostname, config.external_domain), "echo confirmed"], env=env)
+        result = subprocess.check_output(ssh.SSH_BASE + ["-i", keypath, "root@%s.%s" % (hostname, config.external_domain), "echo confirmed"], env=env)
```
krawthekrow: Minor: can use ssh.ssh_get_login. Also "echo confirmed" should be split into "echo" and "confirmed".

celskeggs: I can't use it directly, because I have a hostname, not a node -- but that just requires a bit of tweaking.

celskeggs:
Tested. Everything seems to work. Merging now.

Issue #172, opened 2017-12-27
Node-specific setup steps should run automatically as nodes are admitted to the cluster


celskeggs:
Instead of these steps being performed manually:

    $ spire seq core
    $ spire seq registry
    $ spire seq flannel
    $ spire seq dns-addon

We should be able to just set a configuration option in setup.yaml:

    include-homeworld-registry: true
    
(to say that, yes, the registry *should* be included.)

And then everything else should occur automatically due to nodes being installed and automatic scripts being run.

This will accomplish two things:

 * Easier to set up a cluster
 * Allow adding nodes to an existing cluster.

celskeggs:
The basic commands that need equivalents, which are the unpacking of the `seq` commands from above:

    $ spire infra install-packages
    $ spire setup services
    $ spire verify etcd
    $ spire verify kubernetes
       (the following three commands only if a registry is being set up)
    $ spire setup dns-bootstrap
    $ spire setup bootstrap-registry
    $ spire verify aci-pull
       (end)
    $ spire deploy flannel
    $ spire verify flannel-run
    $ spire verify flannel-ping
    $ spire deploy dns-addon
    $ spire verify dns-addon-run
    $ spire verify dns-addon-ping

There are a few categories of these commands:

  * verification commands, which should probably be converted to monitoring scripts, so that instead of running these directly, it's more of a "let's watch a dashboard to see if all the lights turn green" process
  * kubernetes deployment commands, which might need to be kicked off by a single manual command.
  * configuration that should be able to be done by the ISO, like setting up dns bootstrapping and installing most packages. (homeworld-services can be installed everywhere, since it doesn't have any install-triggered scripts, but the bootstrap registry package can't.)
  * commands that are supervisor-only, and should be able to be performed while setting up the supervisor node (i.e. `spire setup bootstrap-registry`, installing the packages for the bootstrap registry on the supervisor node)
  * commands that might require more complicated machinery to automatically perform, like `spire setup services`

celskeggs:
New ordering (I'm ignoring registry setup checking, because honestly it's not worth the additional effort for that at this time):

      # equivalents will be moved to during ISO installation
    $ spire infra install-packages  (for everything except homeworld-bootstrap-registry)
    $ spire setup dns-bootstrap
      # done during supervisor setup
    $ spire infra install-packages  (for homeworld-bootstrap-registry)
    $ spire setup bootstrap-registry
      # needs more complicated replacement
    $ spire setup services
[...]

PR #178, opened 2017-12-28
Slightly faster deployment

celskeggs:
This should hopefully be easy to review. Not yet tested, but will be shortly.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/seq.py
+++ b/building/build-debs/homeworld-admin-tools/src/seq.py
@@ -1,55 +1,37 @@
-import command
 import time
+
+import access
+import command
 import deploy
 import infra
-import access
-import configuration
 import setup
 import verify
 
 
 def sequence_keysystem(ops: setup.Operations) -> None:
-    # ** Configure the supervisor keyserver **
-
-    # spire setup keyserver
-    setup.setup_keyserver(ops)
-    # spire verify keystatics
+    ops.add_subcommand(setup.setup_keyserver)
     ops.add_operation("verify that keyserver static files can be fetched",
         iterative_verifier(verify.check_keystatics, 10.0))
-
-    # ** Admit the supervisor node to the cluster **
-
-    # spire setup self-admit
-    setup.admit_keyserver(ops)
-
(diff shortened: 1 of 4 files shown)


building/build-debs/homeworld-admin-tools/src/seq.py
```diff
@@ -62,45 +64,84 @@ def ver():
...
+    print("    $ spire deploy dns-addon")
     ops.add_operation("deploy or update dns-addon", lambda: deploy.launch_spec("dns-addon.yaml"))
+    print("  to verify dns-addon:")
+    print("    $ spire verify dns-addon-run")
     ops.add_operation("verify that dns-addon is online", iterative_verifier(verify.check_dns_kubeinfo, 60.0))
+    print("    $ spire verify dns-addon-query")
+    ops.add_operation("verify that dns-addon is functioning", verify.check_dns_function)
+
+
+def sequence_addons(ops: setup.Operations) -> None:
```
krawthekrow: Should reuse sequence_flannel and sequence_dns_addon as in sequence_supervisor.

celskeggs: No -- that's purposeful. They should be interleaved for the sake of spending less time waiting for deployment.

krawthekrow: Alright, if that's intentional. It's minor anyway if the verification steps will ultimately be isolated as per #172.

celskeggs:
I've provided a proposed solution. It's not quite right, given that it involves doing some hacky things with storing attributes on functions, but it seems to work, and I think the model is otherwise good. Thoughts?

celskeggs:
Ah, woops, I accidentally merged the first half of this PR. Hopefully that works out, since that part was approved and seemed uncontentious. The contentious part is still here in this PR.

(I would roll it back, but I figure it's a lesser evil to leave it as-is than to rewrite the history of the master branch.)

celskeggs:
Because we don't currently have infrastructure for handling flags in spire (#181), I'll put those off until later. Other changes put in #182.

I'm going to test this, and then merge it if it passes.

Issue #159, opened 2017-12-17
spire access ssh should complain more loudly if ssh-add fails


ichung:
Currently it's easy to miss the error.

krawthekrow:
Turns out that this does not actually fix the error I had encountered, since ssh-add exits with return code 0. The problem was that [gnome-keyring masquerades as ssh-agent](https://bugzilla.gnome.org/show_bug.cgi?id=697464) but does not support certificates. Until [this is fixed](https://bugzilla.gnome.org/show_bug.cgi?id=775981), it might be worthwhile to detect this case by searching for "add failed" in the ssh-add output. At the same time, the line "or, perhaps your agent is broken; consider killing it and launching a new one." could be removed since it wouldn't show when this particular problem occurs.

celskeggs:
Since you have the ability to reproduce this bug, and I don't, can you take care of resolving this correctly?

PR #183, opened 2017-12-28
Handle error with ssh-add when gnome-keyring is used

krawthekrow:
As per #159. It's a rather ugly way to handle the bug, but then it's a rather ugly bug.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/access.py
+++ b/building/build-debs/homeworld-admin-tools/src/access.py
@@ -74,9 +74,17 @@ def access_ssh(add_to_agent=False):
     print("===== ^ CERTIFICATE DETAILS ^ =====")
     if add_to_agent:
         # TODO: clear old identities
-        if subprocess.call(["ssh-add", "--", keypath]) != 0:
-            fail_hint = "ssh-add returned non-zero exit code. do you have a ssh-agent?\n" \
-                "or, perhaps your agent is broken; consider killing it and launching a new one."
+        try:
+            ssh_add_output = subprocess.check_output(["ssh-add", "--", keypath], stderr=subprocess.STDOUT)
+            # if the user is using gnome, gnome-keyring might
+            # masquerade as ssh-agent and provide a zero exit
+            # code despite failing to add the certificate
+            if b"add failed" in ssh_add_output:
+                fail_hint = "do you have an ssh-agent?\n" \
+                    "(gnome-keyring does not count)"
+                command.fail("*** ssh-add failed! ***", fail_hint)
+        except subprocess.CalledProcessError as e:
+            fail_hint = "ssh-add returned non-zero exit code. do you have an ssh-agent?"
             command.fail("*** ssh-add failed! ***", fail_hint)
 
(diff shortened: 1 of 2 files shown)


building/build-debs/homeworld-admin-tools/src/access.py
```diff
@@ -66,11 +66,17 @@ def access_ssh(add_to_agent=False):
...
-            print("ssh-add returned non-zero exit code. do you have a ssh-agent?")
-            print("or, perhaps your agent is broken; consider killing it and launching a new one.")
-            command.fail("*** ssh-add failed! ***")
-
+        try:
+            ssh_add_output = subprocess.check_output(["ssh-add", "--", keypath], stderr=subprocess.STDOUT)
+            # if the user is using gnome, gnome-keyring might
+            # masquerade as ssh-agent and provide a zero exit
+            # code despite failing to add the certificate
+            if "add failed".encode('utf-8') in ssh_add_output:
```
celskeggs: Do we need to specify the codec? I thought it defaulted to utf-8.

building/build-debs/homeworld-admin-tools/src/access.py
```diff
@@ -66,11 +66,17 @@ def access_ssh(add_to_agent=False):
...
+                fail_hint = "do you have a ssh-agent?"
```
celskeggs: Maybe include "(gnome-keyring does not count)" or something in the hint?

building/build-debs/homeworld-admin-tools/src/access.py
krawthekrow: That's correct. Thanks!

building/build-debs/homeworld-admin-tools/src/access.py
krawthekrow: Good idea, since the branch is only gnome-keyring specific.

building/build-debs/homeworld-admin-tools/src/access.py
```diff
@@ -74,9 +74,17 @@ def access_ssh(add_to_agent=False):
...
         # TODO: clear old identities
-        if subprocess.call(["ssh-add", "--", keypath]) != 0:
-            fail_hint = "ssh-add returned non-zero exit code. do you have a ssh-agent?\n" \
-                "or, perhaps your agent is broken; consider killing it and launching a new one."
+        try:
+            ssh_add_output = subprocess.check_output(["ssh-add", "--", keypath], stderr=subprocess.STDOUT)
+            # if the user is using gnome, gnome-keyring might
+            # masquerade as ssh-agent and provide a zero exit
+            # code despite failing to add the certificate
+            if "add failed".encode() in ssh_add_output:
```
celskeggs: Oh, wait, I just realized what you're doing.

You know you can write this as `b"add failed"` instead of `"add failed".encode()`, right?

krawthekrow: Actually.... I didn't :P Thanks, will change.

Issue #155, opened 2017-12-17
Node consoles should display version/deployment info


ichung:
We should have the motd display some sort of version info such as the timestamp at which the ISO was created to aid in checking which version of the cluster is deployed on a server

ichung:
We should include the git commit hash in this message as in #141, since we already have the machinery for baking it into the build.

PR #189, opened 2018-01-09
Add the current Git commit to the ISO

dmssargent:
This PR addresses the new concern in issue #155.

This commit does the following:
* Adds the hash of the current git HEAD to the preseed.cfg file.
* Adds a required option for the Git commit hash used to build the issue to the ISO's postinstall script
* The postinstall script writes a new line to the /etc/issue file of the new install, containing the commit hash used

### DIFF
--- a/building/build-debs/homeworld-admin-tools/resources/postinstall.sh
+++ b/building/build-debs/homeworld-admin-tools/resources/postinstall.sh
@@ -2,6 +2,7 @@
 echo "launching postinstall"
 
 BUILDDATE="$1"
+GIT_HASH="$2"
 
 . /usr/share/debconf/confmodule
 
@@ -10,6 +11,12 @@ set -e -u
 if [ "$BUILDDATE" = "" ]
 then
     echo "invalid build date" 1>&2
+    exit 1
+fi
+
+if [ "$GIT_HASH" = "" ]
+then
+    echo "invalid git hash" 1>&2
     exit 1
 fi
 
@@ -58,6 +65,7 @@ then
 fi
 
 echo "ISO used to install this node generated at: ${BUILDDATE}" >>/target/etc/issue
+echo "Git commit used to build the version: ${GIT_HASH}" >>/target/etc/issue
 echo "SSH host key fingerprints: (as of install)" >>/target/etc/issue
 for x in /target/etc/ssh/ssh_host_*.pub
 do
--- a/building/build-debs/homeworld-admin-tools/src/iso.py
+++ b/building/build-debs/homeworld-admin-tools/src/iso.py
@@ -10,6 +10,7 @@ import subprocess
 import util
 import packages
 import keycrypt
+from version import get_git_version
 
 PACKAGES = ("homeworld-apt-setup", "homeworld-knc", "homeworld-keysystem")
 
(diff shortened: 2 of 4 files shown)


building/build-debs/homeworld-admin-tools/src/iso.py
```diff
@@ -69,9 +69,12 @@ def gen_iso(iso_image, authorized_key, cdpack=None):
         preseeded = resource.get_resource("preseed.cfg.in")
         generated_password = util.pwgen(20)
         creation_time = datetime.datetime.now().isoformat()
+        git_hash = submodule.check_output("git", "rev-parse", "HEAD")
```
celskeggs: Are you sure that `submodule` is the right thing here? I don't have that available on my system, do you mean `subprocess`?

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -2,6 +2,7 @@
 echo "launching postinstall"
 
 BUILDDATE="$1"
+GIT_HASH = "$2"
```
celskeggs: Might as well keep the spacing consistent.

building/build-debs/homeworld-admin-tools/src/iso.py
dmssargent: I do mean, subprocess. And as you can probably tell, I have not actually run this part of the build. 

When is the ISO generated in the Hyades build system?

celskeggs:
The ISO is generated by the `spire iso gen` command. You can look at `docs/deploy.md` for hints on how to run this.

I wasn't in the office at 1300 -- sorry about that; my schedule was posted on our zephyr class last night, and I didn't have time to check GitHub notifications before 1300.

I'll be in the office through at least 1730 today, and will probably be back later tonight.

building/build-debs/homeworld-admin-tools/src/iso.py
```diff
@@ -69,9 +69,12 @@ def gen_iso(iso_image, authorized_key, cdpack=None):
         preseeded = resource.get_resource("preseed.cfg.in")
         generated_password = util.pwgen(20)
         creation_time = datetime.datetime.now().isoformat()
+        git_hash = subprocess.check_output("git", "rev-parse", "HEAD")
```
celskeggs: This is the wrong invocation of check_output. It takes a single list parameter, not a variable number of parameters.

andersk: Also we’re not in the Git repository when this code runs.

celskeggs: Ah, yes. See `version.py` for the correct source of the version information.

dmssargent:
This PR has now been tested, and it works.

building/build-debs/homeworld-admin-tools/src/iso.py
```diff
@@ -69,9 +70,12 @@ def gen_iso(iso_image, authorized_key, cdpack=None):
         preseeded = resource.get_resource("preseed.cfg.in")
         generated_password = util.pwgen(20)
         creation_time = datetime.datetime.now().isoformat()
+        git_hash = get_resource("GIT_VERSION").decode().rstrip().encode()
```
celskeggs: It would probably be worth adding a function to `version.py` that gets the current version, because we don't want to duplicate that logic.

Issue #168, opened 2017-12-18
Have multiple apt repositories for development


celskeggs:
Currently, we only have one, which gets used both for mainline packages and testing packages. We need an easy way to have multiple and switch between them.

krawthekrow:
Instead of multiple repositories, we could use multiple suites in the same repository:

- [ ] Currently homeworld-apt-setup, homeworld-knc and homeworld-keysystem are manually downloaded and integrated into the ISO by Spire, and installed at post-install. We could do this only for homeworld-apt-setup, and do apt-get update/upgrade/install during post-install. (iso.py, postinstall.sh)
- [ ] Add a switch to select which suite to install from in infra install-packages and affected seq commands. (infra.py, seq.py)
- [ ] Add a switch for upload-debs/rebuild.sh to specify which suite we're building for.
- [ ] Create the testing suite.
- [ ] Add the testing suite to homeworld.sources.

Does this sound like a reasonable approach?

andersk:
I don’t think you want testing to require bits on the same repository that serves prod—that’s going to make development inconvenient and might lead to accidents on prod.

celskeggs:
Agreed with andersk. We probably want to have actual different `apt` repositories, not just different suites.

Also, having just "prod" and "testing" would miss the point. The point is to be able to have arbitrary numbers of repositories -- "prod", "mtheng-1", "cela-trying-out-new-etcd", et cetera.

krawthekrow:
Including packages other than apt-setup in the ISO would require having to set the apt repository source in two places -- in homeworld-apt-setup and in Spire. Here's a proposal for multiple apt repositories:

- [ ] Create and document a process flow for creating a personal homeworld-apt-setup deb.
- [ ] As above, only copy in homeworld-apt-setup when generating the ISO, and update/upgrade/install in post-install.
- [ ] Add a switch to iso.py that chooses whether to download the prod homeworld-apt-setup deb or use the locally built one.
- [ ] Add configuration for upload-debs/rebuild.sh.

celskeggs:
I would consider the following set of changes:

- [ ] Split out the `/building/build-debs/binaries/` folder into multiple variants, one for each apt branch.
- [ ] Require specifying which apt branch we're building for, whenever we build a package. This lets the `homeworld-apt-setup` package and the `homeworld-admin-tools` package know which apt branch is in use.
[...]

PR #201, opened 2018-01-10
Make apt repository configurable

krawthekrow:
Solution for #168.

This solution uses environment variables to propagate the chosen apt repository throughout the building process, and uses a lookup table to find the associated signing key. It requires the developer to have the repository signing key in his GPG keyring. Reprepro updates is used to implement the default repository; this also requires the developer to have the default repository's public key in his GPG keyring (gpg2 specifically, since that's what reprepro uses). ~It currently points to the main homeworld repository, but I'd work on making that configurable.~


building/build-debs/common/package-build-helpers.sh
```diff
@@ -7,9 +7,23 @@ then
 	exit 1
 fi
 
+if [ -z "${APT_REPOSITORY:-}" ]
+then
+	echo "Need to specify apt repository:"
+	echo "\$ export APT_REPOSITORY=<username>/<branch>"
```
celskeggs: You can use single quotes for this, instead of escaping.

building/build-debs/homeworld-admin-tools/src/packages.py
```diff
@@ -149,5 +148,11 @@ def verified_download_full(package_list: tuple) -> dict:
     """Download all the packages from the specified list from the main repository, including verifying them.
 
     Returns a mapping of {package_name: (short_filename, package_bytes), ...}"""
-    verified_info = download_and_verify_package_list(REPO_URL)
-    return {package_name: download_package(package_name, verified_info) for package_name in package_list}
+    apt_repository = resource.get_resource("APT_REPOSITORY").decode().rstrip()
+    apt_url = "http://web.mit.edu/hyades/apt/" + apt_repository
```
celskeggs: The base URL should still be a constant, instead of inline.

building/build-debs/homeworld-apt-setup/build-package.sh
```diff
@@ -2,4 +2,8 @@
 set -eu
 source ../common/package-build-helpers.sh
 
+sed "s;\\\${APT_REPOSITORY};${APT_REPOSITORY};g" < homeworld.sources.in > homeworld.sources
```
celskeggs: Might be worth doing this in a way that doesn't require that a specific character is not present in the value of ${APT_REPOSITORY} -- or at least will properly error out in that case.

building/upload-debs/conf/updates
```diff
@@ -0,0 +1,4 @@
+Name: homeworld
+Components: main
+Method: http://web.mit.edu/hyades/debian/
+VerifyRelease: 9EE78588
```
celskeggs: We should never use an eight-character hash to verify anything, if we can help it. Can we use a 16-character hash here?

building/upload-debs/rebuild.sh
```diff
@@ -1,7 +1,25 @@
...
+
+sed "s;\\\${APT_SIGNING_KEY};${APT_SIGNING_KEY};g" < conf/distributions.in > conf/distributions
+
 cd "$(dirname "$0")"
-reprepro -Vb . includedeb homeworld ../build-debs/binaries/homeworld-*.deb
-rsync -av --progress --delete-delay ./dists ./pool /mit/hyades/debian/
+gpg --import default-repo-signing-key.gpg
+reprepro -Vb . update
+reprepro -Vb . includedeb homeworld ../build-debs/binaries/${APT_REPOSITORY}/homeworld-*.deb
+rsync -av --progress --delete-delay ./dists ./pool /mit/hyades/apt/${APT_REPOSITORY}/
```
celskeggs: You should make sure that you properly quote any substitutions like this.

building/upload-debs/rebuild.sh
```diff
@@ -1,7 +1,25 @@
...
+	exit 1
+fi
+
+sed "s;\\\${APT_SIGNING_KEY};${APT_SIGNING_KEY};g" < conf/distributions.in > conf/distributions
+
 cd "$(dirname "$0")"
-reprepro -Vb . includedeb homeworld ../build-debs/binaries/homeworld-*.deb
-rsync -av --progress --delete-delay ./dists ./pool /mit/hyades/debian/
+gpg --import default-repo-signing-key.gpg
+reprepro -Vb . update
```
celskeggs: Does `update` correctly handle the situation where we're trying to publish to the base repository?

building/upload-debs/rebuild.sh
```diff
@@ -1,7 +1,25 @@
...
+	echo "\$ export APT_SIGNING_KEY=<key identifier>"
+	exit 1
+fi
+
+sed "s;\\\${APT_SIGNING_KEY};${APT_SIGNING_KEY};g" < conf/distributions.in > conf/distributions
+
 cd "$(dirname "$0")"
-reprepro -Vb . includedeb homeworld ../build-debs/binaries/homeworld-*.deb
-rsync -av --progress --delete-delay ./dists ./pool /mit/hyades/debian/
+gpg --import default-repo-signing-key.gpg
```
celskeggs: I'm wary about automatically changing the state of the user's GPG keyring. Can you comment on whether this is necessary?

building/build-debs/homeworld-admin-tools/inner-build.sh
```diff
@@ -18,6 +18,10 @@ then
     echo >>src/resources/GIT_VERSION
 fi
 
+echo -n "$APT_REPOSITORY" >>src/resources/APT_REPOSITORY
```
celskeggs: Is there a reason to have this file lack a trailing newline?

building/upload-debs/conf/updates
```diff
@@ -0,0 +1,4 @@
+Name: homeworld
+Components: main
+Method: http://web.mit.edu/hyades/debian/
```
celskeggs: We probably want to change the upstream branch from `/hyades/debian/` to `/hyades/apt/root/master/` or something. Currently, the modified rebuild.sh script can't actually publish to the upstream branch...

building/build-debs/common/package-build-helpers.sh
```diff
@@ -7,9 +7,23 @@ then
...
+then
+	echo "Need to specify apt signing key:"
+	echo "\$ export APT_SIGNING_KEY=<key identifier>"
+	exit 1
+fi
+
 DVERSION="$(head -n 1 debian/changelog | cut -d '(' -f 2 | cut -d ')' -f 1)"
 VERSION="$(echo "$DVERSION" | cut -d '-' -f 1)"
-BIN=../binaries
+BIN=../binaries/${APT_REPOSITORY}
```
celskeggs: This substitution should probably be quoted.

building/upload-debs/conf/updates
krawthekrow: Configuring upstream is pretty much just changing this updates file; I don't think it's necessary to integrate it into our build process. In fact, if I understand reprepro correctly, you can define multiple upstream repositories here under different names, then it'd just be a matter of changing the name in the "Update:" field in conf/distributions to decide which one to use.

Could you move /hyades/debian/ to /hyades/apt/root/master/? (It's currently empty.) I don't have the bits for that.

building/upload-debs/rebuild.sh
krawthekrow: I'm also uncomfortable with this part. However, reprepro doesn't seem to have a mechanism for using a key file directly. Looking at the man pages, though, it looks like I can specify a temporary GnuPG home directory for reprepro to use; I'll try doing that.

building/upload-debs/rebuild.sh
krawthekrow: In theory, it should -- it only pulls updates if the version number is less than your currently built version, so you would have the updated debs in your repository when you rsync them over. It would be better to test this though -- could you do it just before merge? I don't have the bits for that

building/upload-debs/rebuild.sh
krawthekrow: My mistake, will fix.

building/build-debs/homeworld-apt-setup/build-package.sh
krawthekrow: It would be nice if Bash had a cleaner way to do substitutions in files. Is this a security concern, or are we just handling the possibility of having special characters in future apt branches? I could use something from https://stackoverflow.com/a/2705678/1263980, but it might be easier (and better) to just limit the characters allowed in apt branch names, like to [A-Za-z0-9_-].

building/build-debs/homeworld-admin-tools/inner-build.sh
krawthekrow: There isn't, but only because rstrip is done on the Python side. I'll remove it the flag here.

building/build-debs/common/package-build-helpers.sh
krawthekrow: Thanks! Forgot you could do that.

building/build-debs/homeworld-apt-setup/build-package.sh
celskeggs: Just limiting the characters sounds good to me.

building/upload-debs/conf/updates
celskeggs: I've moved the folder, and left a symlink from /hyades/debian.

Your thoughts on configurable upstreams seem reasonable. Good enough for me.

building/upload-debs/rebuild.sh
celskeggs: Both of those work. Using a temporary GnuPG directory is good. Verifying that the key is in the keyring and telling the developer to import it if not is also good.

building/upload-debs/rebuild.sh
celskeggs: Yeah, I can test this once we're closer to merge.

Have we confirmed that `update` verifies signatures properly?

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -11,5 +11,10 @@ git rev-parse HEAD | tr -d '\n' >resources/GIT_VERSION
 [[ -n "$(git status --porcelain)" ]] && echo -n "-dirty" >>resources/GIT_VERSION
 echo >>resources/GIT_VERSION
 
+echo "${HOMEWORLD_APT_BRANCH}" >>resources/APT_BRANCH
+
+HOMEWORLD_APT_SIGNING_KEY="$(cd ../.. && source get-apt-signing-key.sh)"
```
celskeggs: Why does this need to be sourced? Why do we need to change directories to use it?

Any reason not to make it a function instead of an entire script?

building/build-debs/homeworld-admin-tools/build-package.sh
```diff
@@ -11,5 +11,10 @@ git rev-parse HEAD | tr -d '\n' >resources/GIT_VERSION
...
+gpg --export "${HOMEWORLD_APT_SIGNING_KEY}" >>resources/homeworld-archive-keyring.gpg
```
celskeggs: Maybe we want to clean up this file automatically, just like GIT_VERSION?

building/build-debs/homeworld-admin-tools/debian/rules
```diff
@@ -17,7 +17,7 @@ include /usr/share/dpkg/default.mk
 #export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
 
 override_dh_auto_build:
-	VERSION="${DEB_VERSION_UPSTREAM}" ./inner-build.sh
+	VERSION="${DEB_VERSION_UPSTREAM}" HOMEWORLD_APT_BRANCH="${HOMEWORLD_APT_BRANCH}" ./inner-build.sh
```
celskeggs: Huh. Why is this necessary?

building/build-debs/homeworld-admin-tools/src/packages.py
```diff
@@ -149,5 +150,11 @@ def verified_download_full(package_list: tuple) -> dict:
...
-    verified_info = download_and_verify_package_list(REPO_URL)
-    return {package_name: download_package(package_name, verified_info) for package_name in package_list}
+    apt_branch = resource.get_resource("APT_BRANCH").decode().rstrip()
+    apt_url = APT_REPO_BASE + apt_branch
+    try:
+        verified_info = download_and_verify_package_list(apt_url)
+        return {package_name: download_package(package_name, verified_info) for package_name in package_list}
+    except urllib.error.HTTPError:
+        command.fail("unable to access apt repository",
+            "do you have an apt repository at %s?" % apt_url)
```
celskeggs: Might be worth using consistent terminology: apt branch, not apt repository.

building/get-apt-signing-key.sh
```diff
@@ -0,0 +1,21 @@
+while read branch key; do
+	if [ "${branch}" == "${HOMEWORLD_APT_BRANCH}" ] || [ "${branch}" == "$(echo ${HOMEWORLD_APT_BRANCH} | cut -d '/' -f 1)" ]; then
+		HOMEWORLD_APT_SIGNING_KEY="${key}"
+		break
+	fi
+done < signing-keys
```
celskeggs: For the sake of readability, I think people usually wrap constructs in parentheses if they're going to have things redirected into them afterwards?

i.e.:

    (for x in y; do
        <code>
    done) > output-file

building/get-apt-signing-key.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ -z "${HOMEWORLD_APT_SIGNING_KEY:-}" ]
+then
+	echo 'apt branch not found in signing-keys'
```
celskeggs: Since the result is normally echoed, it might be worth making sure that errors are sent to stderr? (`1>&2`)

building/validate-apt-branch.sh
```diff
@@ -0,0 +1,14 @@
...
+	exit 1
+fi
+
+apt_branch_regex='^[0-9a-zA-Z_-]+\/[0-9a-zA-Z_-]+$'
+if ! [[ "${HOMEWORLD_APT_BRANCH}" =~ $apt_branch_regex ]]
+then
+	echo 'Apt branch invalid. Should be of the form <username>/<branch>.'
+	echo 'Allowed characters: [0-9a-zA-Z_-]'
+	exit 1
+fi
```
celskeggs: I might consider turning get-apt-signing-key.sh into a function in this file.

docs/build.md
```diff
@@ -17,7 +17,26 @@ That will let you build everything as described here.  If you also want to be ab
  * python3-yaml
  * zip
 
+# Create a personal apt repository
+
+Set the apt branch environment variable (you will have to do this every session):
+
+    $ export HOMEWORLD_APT_BRANCH=<username>/<branch>
+
+Do `gpg --list-keys --keyid-format long` to find the key you would like to sign your repository with. Add an entry to `building/signing-keys`:
```
celskeggs: Perhaps suggest generating a key, and specifically discourage from using a regular long-lived PGP key?

docs/build.md
```diff
@@ -17,7 +17,26 @@ That will let you build everything as described here.  If you also want to be ab
...
+
+    $ echo "${HOMEWORLD_APT_BRANCH} <key-id>" >> building/signing-keys
+
+Import the default apt repository signing key:
+
+    $ cd building/upload-debs
+    $ gpg --import default-repo-signing-key.gpg
+    $ cd ../..
```
celskeggs: You can just have them import a relative path, instead of cding there.

building/validate-apt-branch.sh
```diff
@@ -0,0 +1,14 @@
+if [ -z "${HOMEWORLD_APT_BRANCH:-}" ]
```
celskeggs: It might be worth sticking these two script in a subdirectory of `/building/`, as directly under `/building/` is a strange place to stick scripts.

building/build-debs/homeworld-admin-tools/debian/rules
krawthekrow: Oops, forgot to remove that.

building/setup-apt-branch/setup-apt-branch.sh
```diff
@@ -0,0 +1,39 @@
...
+    done < `dirname $BASH_SOURCE`/signing-keys
+
+    if [ -z "${HOMEWORLD_APT_SIGNING_KEY:-}" ]
+    then
+        echo 'error: apt branch not found in signing-keys' >&2
+        exit 1
+    fi
+
+    apt_signing_key_regex='^[0-9a-zA-Z]+$'
+    if ! [[ "${HOMEWORLD_APT_SIGNING_KEY}" =~ $apt_signing_key_regex ]]
```
celskeggs: Is there a reason why you're defining a variable, just to use it exactly once?

building/setup-apt-branch/setup-apt-branch.sh
```diff
@@ -0,0 +1,39 @@
...
+    then
+        echo 'error: apt signing key invalid' >&2
+        exit 1
```
celskeggs: Have you confirmed that this error checking works?

docs/build.md
```diff
@@ -42,6 +63,8 @@ To upload packages, ONLY IF YOU ARE SUPPOSED TO RELEASE YOUR CHANGES:
 
     $ cd building/upload-debs
     $ aklog sipb    # authenticate to the SIPB AFS cell, if necessary
+        # Note that you will need Kerberos tickets
+        # (generate them with kinit) to access AFS.
```
celskeggs: Any reason that one of these is after the 'aklog' and one of these is before?

building/setup-apt-branch/setup-apt-branch.sh
krawthekrow: Yes, I have tested this manually with different inputs.

docs/build.md
krawthekrow: No, I'll fix that.

building/setup-apt-branch/setup-apt-branch.sh
krawthekrow: No. I had problems getting the inline regex to work earlier, but it works now that other bugs have been fixed. I'll change both regexes to inline.

celskeggs:
Needs rebase.

Issue #144, opened 2017-12-05
Confirm that a .pem is public material before sticking it into authorities.tgz


celskeggs:
Just in case.

Original comment: https://github.com/sipb/homeworld/pull/135#discussion_r154534577

In other words: instead of assuming that every `.pem` file is safe to copy as public information, we should read the headers on them to ensure that they contain a PEM certificate and not anything else.

dmssargent:
Just to confirm, you only want .pem files to have just certificate(s) in them and nothing else per what is allowable by the format? Or do you want to ensure the .pem file does not contain any private keys?

For future reference, [RFC 7468](https://tools.ietf.org/html/rfc7468) specifies the encapsulation methods used in the PEM file format, while [RFC 1421](https://tools.ietf.org/html/rfc1421) Section 4.4 considers only the text that is encapsulated to be a PEM message. OpenSSL's [pem.h](https://github.com/openssl/openssl/blob/master/include/openssl/pem.h) can show us what types of messages PEM files can carry.

ichung:
@dmssargent The former - there should be only a certificate and nothing else.

dmssargent:
Thanks @ichung!

dmssargent:
If a file is found that has the extension of ".pem" but contains >1 certificates or contains other things (like a private key), what should my code do?

celskeggs:
Throw an exception. Probably just use command.fail for that.

PR #205, opened 2018-01-12
Verify PEM files contain only 1 certificate

dmssargent:
This PR addresses issue #144.

The PEM files used to create the authorities.tgz should only be certificates and public material. This PR adds additional validation to the PEM files to verify that they do contain certificates and only certificates. Otherwise, `spire authority gen` will fail with an error message

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/authority.py
+++ b/building/build-debs/homeworld-admin-tools/src/authority.py
@@ -23,7 +23,25 @@ def name_for_decrypted_file(name_of_encrypted_file):
     if name_of_encrypted_file.endswith(ENCRYPTED_EXTENSION):
         return name_of_encrypted_file[:-len(ENCRYPTED_EXTENSION)]
     raise ValueError("Filename " + name_of_encrypted_file + " does not have expected suffix '" + ENCRYPTED_EXTENSION + "'.")
-    
+   
+def validate_pem_file(full_file_name: str) -> bool:
+    with open(full_file_name) as pem_file:
+        has_certificate = False
+        for line in pem_file:
+            # PEM data that is a certificate should have the following header
+            if line.startswith("-----BEGIN CERTIFICATE-----"):
+                if has_certificate:
+                    command.fail("pem file \"" + file_name + "\" should only contain 1 certificate, but it contains additional certificates!")
+                has_certificate = True
+            # The file should not have other types of PEM data
+            elif line.startswith("-----BEGIN "):
+                command.fail("pem file \"" + file_name + "\" contains non-certificate PEM data!")
+        # The file should have had a certificate
+        if not has_certificate:
+            command.fail("pem file \"" + file_name + "\" does not contain any PEM certificates!")
+
+        return True
+
 def generate() -> None:
     authorities = get_targz_path(check_exists=False)
     if os.path.exists(authorities):
@@ -47,7 +65,7 @@ def generate() -> None:
         cryptdir = os.path.join(d, "cryptdir")
         os.mkdir(cryptdir)
         for filename in os.listdir(certdir):
-            if filename.endswith(".pub") or filename.endswith(".pem"):
+            if filename.endswith(".pub") or (filename.endswith(".pem") and validate_pem_file(os.path.join(certdir, filename))):
                 # public keys; copy over without encryption
                 util.copy(os.path.join(certdir, filename), os.path.join(cryptdir, filename))
             else:


building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -23,7 +23,24 @@ def name_for_decrypted_file(name_of_encrypted_file):
     if name_of_encrypted_file.endswith(ENCRYPTED_EXTENSION):
         return name_of_encrypted_file[:-len(ENCRYPTED_EXTENSION)]
     raise ValueError("Filename " + name_of_encrypted_file + " does not have expected suffix '" + ENCRYPTED_EXTENSION + "'.")
-    
+   
+def validate_pem_file(dir_path: str, file_name: str) -> None:
+    full_file_name = os.path.join(dir_path, file_name)
```
celskeggs: If all this needs is a filename, it should be passed in as a filename. There's no reason for this function to join the filename itself.

If you want a short version of the name for a file, use os.path.basename to extract it.

building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -23,7 +23,24 @@ def name_for_decrypted_file(name_of_encrypted_file):
...
+        has_certificate = False
+        for line in pem_file:
+            # PEM data that is a certificate should have the following header
+            if line.startswith("-----BEGIN CERTIFICATE-----"):
+                if has_certificate:
+                    command.fail("The PEM file \"" + file_name + "\"should only contain 1 certificate, but it contains additional certificates!")
+                has_certificate = True
+            # The file should not have other types of PEM data
+            elif line.startswith("-----BEGIN "):
+                command.fail("The PEM file \"" + file_name + "\"should only contain 1 certificate, but it contains additional PEM data!")
```
celskeggs: Error message is unclear. Saying "should only contain 1" implies that you've already found one, even though this can occur when that's not the case.

building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -23,7 +23,24 @@ def name_for_decrypted_file(name_of_encrypted_file):
...
+        # The file should have had a certificate
+        if not has_certificate:
+                command.fail("The PEM file \"" + file_name + "\"should only contain 1 certificate, but it does not contain any PEM data!")
```
celskeggs: You can be more concise here by just skipping to "does not contain" instead of talking about "should only contain 1 certificate"

building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -23,7 +23,24 @@ def name_for_decrypted_file(name_of_encrypted_file):
...
+
```
celskeggs: Your indentation looks odd here.

building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -23,7 +23,24 @@ def name_for_decrypted_file(name_of_encrypted_file):
...
+   
+def validate_pem_file(dir_path: str, file_name: str) -> None:
+    full_file_name = os.path.join(dir_path, file_name)
+    with open(full_file_name) as pem_file:
+        has_certificate = False
+        for line in pem_file:
+            # PEM data that is a certificate should have the following header
+            if line.startswith("-----BEGIN CERTIFICATE-----"):
+                if has_certificate:
+                    command.fail("The PEM file \"" + file_name + "\"should only contain 1 certificate, but it contains additional certificates!")
```
celskeggs: Error messages through command.fail should start with a lowercase letter.

building/build-debs/homeworld-admin-tools/src/authority.py
```diff
@@ -47,7 +64,7 @@ def generate() -> None:
         cryptdir = os.path.join(d, "cryptdir")
         os.mkdir(cryptdir)
         for filename in os.listdir(certdir):
-            if filename.endswith(".pub") or filename.endswith(".pem"):
+            if filename.endswith(".pub") or (filename.endswith(".pem") and validate_pem_file(certdir, filename)):
```
celskeggs: You're using validate_pem_file here as a boolean, which is declared not to return anything.

Did you test this code? I'm not quite sure why this would work.

dmssargent: It works, although incorrectly.

For me:
```bash
$ python3 --version
Python 3.6.3
```

The following code prints 0.
```python
if True and None:
    print(1)
else:
    print(0)
```

celskeggs: "Working, although incorrectly" is not a kind of "working."

(i.e. yes, you *can* use a None as a boolean. It evaluates to false. But I don't think you mean to do that.)

dmssargent:
I pushed a new changes awhile ago (a few days ago), but I don't think it sent any notifications.

celskeggs:
The usual technique is to just assign the issue back to me. I'll take a look when I have the chance.

dmssargent:
Oh, okay. Thanks!

Issue #210, opened 2018-01-17
include apt branch in spire version output


celskeggs:
It would be good to have an easy way to check the actual apt branch in use by spire.

Issue #209, opened 2018-01-17
apt branching requires incrementing versions of apt-setup and admin-tools


celskeggs:
This is because reprepro won't allow us to replace the packages from root/master with our own same-versioned packages.

Issue #208, opened 2018-01-17
apt branch pools/dists/db/lists are not distinct


celskeggs:
This means that switching between apt branches will cross-contaminate the repository state.

Workaround: `rm -rf db dists lists pool` whenever you change branches.

PR #214, opened 2018-01-18
Clean up apt branching

krawthekrow:
Solves #208, #209, #210 and #213.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/version.py
+++ b/building/build-debs/homeworld-admin-tools/src/version.py
@@ -7,12 +7,19 @@ def get_git_version():
     return get_resource("GIT_VERSION").decode().rstrip()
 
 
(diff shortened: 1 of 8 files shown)


building/build-debs/homeworld-admin-tools/src/version.py
```diff
@@ -14,5 +14,8 @@ def display_version():
     git_version = get_git_version()
     print("Git commit hash:", git_version)
 
+    apt_branch = get_resource("APT_BRANCH").decode().rstrip()
+    print("Apt branch:", apt_branch)
+
```
celskeggs: Perhaps this should probably be refactored into a separate function, since it's used more than once?

krawthekrow: Good idea.

Issue #132, opened 2017-12-03
Bootstrap tokens should include a check code


ichung:
If you typo the code, you currently don't find out until it doesn't work later.  We should add a check code to prevent typos.

celskeggs:
@guillde: do you expect to be able to resolve this issue in the near future? If not, we should unassign you so that it's freed up.

(We can also find you an easier issue, like #162 or #163.)

guillde:
You can unassign the issue. And do not worry about finding me another issue.


________________________________

celskeggs:
Solved by #216.

PR #216, opened 2018-01-19
Add code to have the user re-confirm the bootstrap token

dmssargent:
This PR addresses #132 

Currently, the user only has to type the bootstrap token once, which is quite vulnerable to someone making a typo. This PR forces the bootstrap token to typed at least twice, if the tokens do not match then the installer will continue to prompt for the bootstrap tokens until they do match.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/resources/postinstall.sh
+++ b/building/build-debs/homeworld-admin-tools/resources/postinstall.sh
@@ -45,6 +45,10 @@ Template: homeworld/asktoken
 Type: string
 Description: Enter the bootstrap token for this server.
 
+Template: homeworld/tokeninvalid
+Type: note
+Description: Invalid token! Please check the token for typos. Press Enter to try again.
+
 Template: homeworld/title
 Type: text
 Description: Configuring keysystem...
@@ -52,12 +56,31 @@ EOF
 
 debconf-loadtemplate homeworld /tmp/token.template
 
-db_settitle homeworld/title
+is_invalid_token () {
+    TOKEN="${1%??}"
+    TOKEN_PROVIDED_HASH=$(echo -n $1 | tail -c 2)
+    TOKEN_ACTUAL_HASH=$(echo -n "$TOKEN" | /target/usr/bin/openssl dgst -sha256 -binary | /target/usr/bin/base64 | cut -c -2)
 
+    if [ "$TOKEN_PROVIDED_HASH" != "$TOKEN_ACTUAL_HASH" ]; then
+        return 0
+    fi
+
+    return 1
+}   
+
+db_settitle homeworld/title
 db_input critical homeworld/asktoken
 db_go
 
 db_get homeworld/asktoken
+while [ "$RET" != "manual" ] && is_invalid_token $RET; do
+    db_input critical homeworld/tokeninvalid
+    db_go
+
+    db_input critical homeworld/asktoken
+    db_go
+    db_get homeworld/asktoken
+done
 
 if [ "$RET" != "" ]
 then
--- a/building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go
+++ b/building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go
@@ -2,6 +2,7 @@ package scoped
 
 import (
 	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
 	"errors"
 	"sync"
@@ -37,7 +38,10 @@ func generateTokenID() string {
 	if err != nil {
 		panic(err)
 	}
-	return base64.RawStdEncoding.EncodeToString(out)
+
+	hash := base64.RawStdEncoding.EncodeToString(out)
+	hashSha256 := sha256.Sum256([]byte(hash))
+	return hash + base64.RawStdEncoding.EncodeToString(hashSha256[:])[0:2]
 }
 
 func GenerateToken(subject string, duration time.Duration) ScopedToken {
--- a/building/build-debs/homeworld-admin-tools/src/infra.py
+++ b/building/build-debs/homeworld-admin-tools/src/infra.py
@@ -18,10 +18,10 @@ def infra_admit_all() -> None:
         principal = node.hostname + "." + config.external_domain
         token = access.call_keyreq("bootstrap-token", principal, collect=True).decode().strip()
         tokens[node.hostname] = (node.kind, node.ip, token)
-    print("host".center(16, "="), "kind".center(8, "="), "ip".center(14, "="), "token".center(21, "="))
+    print("host".center(16, "="), "kind".center(8, "="), "ip".center(14, "="), "token".center(23, "="))
     for key, (kind, ip, token) in sorted(tokens.items()):
-        print(key.rjust(16), kind.center(8), str(ip).center(14), token.ljust(21))
-    print("host".center(16, "="), "kind".center(8, "="), "ip".center(14, "="), "token".center(21, "="))
+        print(key.rjust(16), kind.center(8), str(ip).center(14), token.ljust(23))
+    print("host".center(16, "="), "kind".center(8, "="), "ip".center(14, "="), "token".center(23, "="))
 
 
 def infra_install_packages(ops: setup.Operations) -> None:
--- a/building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token_test.go
+++ b/building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token_test.go
@@ -1,6 +1,8 @@
 package scoped
 
 import (
+	"crypto/sha256"
+	"encoding/base64"
 	"math"
 	"math/rand"
 	"strings"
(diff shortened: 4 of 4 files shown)


ichung:
I think this adds friction to the process without much benefit.  Why not just add a checksum to the token as previously discussed?On Jan 19, 2018 18:21, David Sargent <notifications@github.com> wrote:This PR addresses #132
Currently, the user only has to type the bootstrap token once, which is quite vulnerable to someone making a typo. This PR forces the bootstrap token to typed at least twice, if the tokens do not match then the installer will continue to prompt for the bootstrap tokens until they do match.



You can view, comment on, or merge this pull request online at:
  https://github.com/sipb/homeworld/pull/216

Commit Summary
Add code to have the user re-confirm the bootstrap token

File Changes

    M
    building/build-debs/homeworld-admin-tools/resources/postinstall.sh
    (42)
  

Patch Links:
https://github.com/sipb/homeworld/pull/216.patchhttps://github.com/sipb/homeworld/pull/216.diff

dmssargent:
I could implement checking some form of checksum check before accepting the code, but I am not familiar with any previous discussions about this subject, so could someone explain what the original opinions were on the subject?

ichung:
The original idea was to append an [error-correcting code](https://en.wikipedia.org/wiki/Forward_error_correction) to the token so that we can detect (or even correct) typos.  The solution of typing in the entire token twice is much less efficient than a well-chosen error-correcting code.

dmssargent:
After a bit of research, would it be okay to just have a normal hashcode appended to the end of the string?

I would be thinking along the lines of generating the hash code for the token in a way similar to how Java implements the hashcode() function for a string. Then the output of that function (int) could be converted to two Base64 chars (if I did my math correctly, that would allow for 8191 possible hashcodes)

Let me know if there is a much simpler, or better way of doing this, because I didn't find any other trivial way to implement this.

celskeggs:
That sounds good, but I'd use [sha256](https://golang.org/pkg/crypto/sha256/#Sum256) instead. (And I think two base64 digits is 64 * 64 = 4096 possibilities?)

dmssargent:
I thought the goal was only one or two more characters, or do you want me to consider only some of the sha256 hash (like I would be for the original hashcode() concept)

Also, you are right 64*64=only 4096 combinations

celskeggs:
That's correct -- you'd use the first 12 bits of the sha256 to form two base64 digits. (Really, you just want to take a hash, base64-encode it, and then take the first two digits.)

dmssargent:
An update...I have a working minimal PoC bash script that does performs these checks, and I have migrated the PoC to postinstall.sh, but I haven't tested the code on the cluster yet, since I have been distracted with a few other projects with more pressing deadlines. 

I will test (and push the changes to this PR) tomorrow for at least the bootstrap token "manual," (if the cluster lock is available), but should I drop by the office for a guide to actual bootstrap tokens?

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -50,23 +51,46 @@ Type: text
...
-debconf-loadtemplate homeworld /tmp/token.template
+cat >/tmp/token-invalid.template <<EOF
+Template: homeworld/invalidtoken
+Type: string
+Description: Invalid token! Please check the token for typos.
 
+Template: homeworld/title
+Type: text
+Description: Configuring keysystem...
+EOF
```
celskeggs: Is it possible to merge these two templates?

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -50,23 +51,46 @@ Type: text
...
+while [ "$RET" != "manual" ] && [ "$TOKEN_PROVIDED_HASH" != "$TOKEN_ACTUAL_HASH" ]; do
+    db_input critical homeworld/invalidtoken
+    db_go
+
+    db_get homeworld/invalidtoken
+
+    TOKEN=${RET%??}
+    TOKEN_PROVIDED_HASH=`echo -n $RET | tail -c 2`
+    TOKEN_ACTUAL_HASH=`echo -n "$TOKEN" | /target/usr/bin/openssl dgst -sha256 -binary | /target/usr/bin/base64 | cut -c -2`
+done
```
celskeggs: It's probably worth reordering this so that it doesn't duplicate any code.

dmssargent:
Where is the code that generates those bootstrap tokens?

celskeggs:
In the keysystem, [here](https://github.com/sipb/homeworld/blob/master/building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go#L34).

dmssargent:
Thanks! Would it been fine to just generate the hash in generateTokenID() function, and append it to the token there? Or is the size of 15 important to something?

From what I can tell the bootstrap token themselves are meaningless, so just adding the additional checksum before the token gets added to the registry should be fine, but then all subsequent requests that uses that token will have to send that full token back. I am also assuming that you want the token to be in the correct format, otherwise Check() should fail?

celskeggs:
That should be fine. The size of 15 is unimportant, and there's no meaningful cost of transmitting two additional bytes.

Since the only tokens that are valid are ones that were previously generated by generateTokenID(), you shouldn't need to explicitly verify the checksum in the keyserver.

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -40,17 +40,34 @@ cp /keyclient-*.yaml /target/etc/homeworld/config/
...
 Type: text
 Description: Configuring keysystem...
 EOF
 
-debconf-loadtemplate homeworld /tmp/token.template
+}
+
+generate_token_template "token" "Enter the bootstrap token for this server."
+generate_token_template "tokeninvalid" "Invalid token! Please check the token for typos."
+debconf-loadtemplate homeworld /tmp/token.template /tmp/tokeninvalid.template
```
celskeggs: What I meant was -- can you have token.template and tokeninvalid.template be merged into one template file?

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -40,17 +40,34 @@ cp /keyclient-*.yaml /target/etc/homeworld/config/
...
+    TOKEN="${1%??}"
+    TOKEN_PROVIDED_HASH=`echo -n $1 | tail -c 2`
+    TOKEN_ACTUAL_HASH=`echo -n "$TOKEN" | /target/usr/bin/openssl dgst -sha256 -binary | /target/usr/bin/base64 | cut -c -2`
+
+    if [ "$TOKEN_PROVIDED_HASH" != "$TOKEN_ACTUAL_HASH" ]; then
+        return 0
+    fi
+
+    return 1
+}
```
celskeggs: The name "validate_token" implies that it returns true when valid and false when invalid, whereas this does the opposite. Perhaps a different name would make this better?

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -40,17 +40,34 @@ cp /keyclient-*.yaml /target/etc/homeworld/config/
...
-debconf-loadtemplate homeworld /tmp/token.template
+}
+
+generate_token_template "token" "Enter the bootstrap token for this server."
+generate_token_template "tokeninvalid" "Invalid token! Please check the token for typos."
+debconf-loadtemplate homeworld /tmp/token.template /tmp/tokeninvalid.template
+
+validate_token () {
+    TOKEN="${1%??}"
+    TOKEN_PROVIDED_HASH=`echo -n $1 | tail -c 2`
```
celskeggs: Optional -- but you might consider using `$(...)` instead of ``` `...` ```

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
```diff
@@ -40,17 +40,34 @@ cp /keyclient-*.yaml /target/etc/homeworld/config/
...
+    TOKEN_ACTUAL_HASH=`echo -n "$TOKEN" | /target/usr/bin/openssl dgst -sha256 -binary | /target/usr/bin/base64 | cut -c -2`
```
celskeggs: It seems a bit weird to be using the binaries installed on the new system to do a computation here. Are these binaries not installed in the ISO's own filesystem?

building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go
```diff
@@ -2,6 +2,7 @@ package scoped
 
 import (
 	"crypto/rand"
+    "crypto/sha256"
```
celskeggs: Indentation. Consider running `gofmt`.

building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go
```diff
@@ -37,7 +38,10 @@ func generateTokenID() string {
 	if err != nil {
 		panic(err)
 	}
-	return base64.RawStdEncoding.EncodeToString(out)
+
+	hash := base64.RawStdEncoding.EncodeToString(out)
+	hashSha256 := sha256.Sum256([]byte(hash))
+	return hash + base64.RawStdEncoding.EncodeToString(hashSha256[:])[0:2]
```
ichung: I suggest adding a test to ensure that this function always generates valid tokens

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
dmssargent: Oh, merged into one template file, apparently I can do that.

building/build-debs/homeworld-admin-tools/resources/postinstall.sh
dmssargent: Nope, they were not preinstalled on the ISO's file system.

celskeggs: Oh well.

building/build-debs/homeworld-keysystem/src/keyserver/token/scoped/token.go
dmssargent: Done

dmssargent:
Not all the way, yet. I will try doing that in about 2-3 hours.

celskeggs:
Also, you'll want to rebase against master before testing.

dmssargent:
Be sure to upgrade the homeworld-keysystem apt package on the homeworld apt branch, or manually deploy the new keysystem package to the supervisor node, so everything installs smoothly. Everything seemed to test out fine, but the first time the token has a typo the transition to the invalid template causes the token to need to be retyped. Do you want me to strip the invalid token template from the postinstall script (it would simplify things, but it may make for a more confusing UX) or do you want to me to preserve the token in the invalid token template (at the cost of more added complexity to the script)?

celskeggs:
Maybe a good solution would be to have an interstitial screen that says `You typed the bootstrap token incorrectly. Press OK to try again.` or something? You can then just reuse the same token template without much confusion.

PR #228, opened 2018-01-22
Convert verification scripts to use prometheus

### LINKED ISSUE #180: Basic monitoring
As part of #172, and as a step towards #16, we need to set up some infrastructure that can monitor different parts of a cluster, either during setup or production.

At some point, this will plug into Prometheus, probably, but we need a broader system, so that we have something that works before we can launch prometheus.

This will monitor elements that include, but are not limited to:

  * That each node is online.
  * ~That each systemd service is running on each node.~ deferred, since this isn't necessary to replace the existing verification scripts.
  * That etcd is healthy.
  * That kubernetes is healthy.
  * That containers can be pulled from the registry.
  * That the core cluster services (flannel, dns-addon) are actively running.
  * That the flannel overlay network is functioning.
  * That DNS is actively accessible (via the dns-addon).
  * That SSH access is possible via cluster auth.

### LINKED ISSUE #195: Convert verification scripts to be based on prometheus monitoring
As part of #180, we're moving the verification scripts to have all of their necessary data be tracked by the monitoring system. Once that's done (#190, #191, #192, #193, #194), we can convert the verification scripts, which is this issue.

### LINKED ISSUE #203: spire seq ssh runs spire verify ssh-with-certs too soon
In some cases, the SSH daemon doesn't have enough time to restart (I think?), leading to this error:

    -- verify ssh access to supervisor  (35/35) --
    Permission denied (publickey).
    command failed: ssh check failed: Command '['ssh', '-o', 'StrictHostKeyChecking=
    yes', '-o', 'ConnectTimeout=1', '-i', '/home/user/dc6-proto1/ssh-key', 'root@egg
    -sandwich.mit.edu', 'echo', 'confirmed']' returned non-zero exit status 255

If I immediately run `spire verify ssh-with-certs` manually, it passes, so I think this is just a tiny timing issue.

celskeggs:
Fixes #195. Fixes #180. Fixes #203.

~There is a known problem that the use of recording rules in prometheus leads to high latency for the verification scripts. This problem will be fixed in a later PR.~


building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -317,6 +317,16 @@ def get_prometheus_yaml() -> str:
     return template.template("prometheus.yaml", kcli)
 
 
+EXPECTED_KUBERNETES_VER = "1.8.0"
+
+
+def get_prometheus_monitor_rules() -> str:
+    config = Config.load_from_project()
+    kcli = {"REGEXVER": EXPECTED_KUBERNETES_VER.replace(".", "[.]"),
+            "REGEXNODES": "|".join(node.hostname for node in config.nodes if node.kind == "master")}
```
krawthekrow: Hostnames could contain special characters, like dots.

celskeggs:
Hold off on reviewing this; my next set of changes deletes enough of this code that it doesn't make sense to review this checkpoint.

celskeggs:
Ready for review again.

Issue #188, opened 2018-01-09
Perform DNS bootstrap setup during ISO installation


celskeggs:
If node should start with DNS bootstrapping turned on, then it shouldn't require a command to run.

Part of #172.

celskeggs:
@minced: how's this going?

celskeggs:
@minced: do you expect to have time to finish this in the near future? If not, we'll need to assign someone else.

celskeggs:
@minced: I remember that you showed me most of the required code for this in person recently, but I haven't heard much; if you won't be able to finish this by the meeting on Monday, can you at least post the code you have somewhere so that someone else can finish it?

minced:
Hi! I'm so sorry I'm late. Code has been posted to the dns-bootstrap-during-iso-install branch. I will attempt to test my changes tonight if it's not too late.

minced:
(Also note: I pushed to origin/dns-bootstrap... instead of minced/dns-bootstrap... at first; origin branch should probably be deleted)

celskeggs:
Merged as part of #269.

PR #229, opened 2018-01-24
Streamline deployment process further

celskeggs:
The four main ideas here:

 * Add a homeworld-autostart system to automatically launch services on each node after they're admitted to the cluster.
 * Change kubernetes spec deployment to occur through a queue.
 * Improve sequencing of keyclient-related events by using systemd-notify for keyclient.
 * Change kube-state-metrics to run on the supervisor node, so that we can monitor the kubernetes cluster even when the cluster is broken or not yet configured.

This almost finishes #172, except for #188. With this PR, I deployed a cluster in under 13 minutes, and it should (in theory) now be possible to add additional members to the cluster without reconfiguring all of the old nodes.


krawthekrow:
Is it a bad idea to have the keysystem run the startup script upon admission, or make admission part of the startup process? Using a timer for this seems a bit kludgy.

celskeggs:
Admission *is* part of the startup process. It is possible for the keyclient to run a script once it's done, but I'd worry that it couples the components too tightly, and has weird edge case problems with things like the autostart script being run as a child process of the keyclient, and potentially causing problems if it waits too long or exits.

It's not a bad idea; I'm not convinced it's unambiguously better than this model.

krawthekrow:
What about using systemd notify to notify systemd when the keyclient is ready to process requests? The other services can then be made to depend on it.

I was concerned that using a timer doesn't make it explicit that the other services depend on this admission process being completed, and could easily lead to timing issues if overused.

celskeggs:
The keyclient doesn't process requests. It performs a set of operations, like generating keys, getting them signed by the keyserver, and downloading files from the keyserver.

You're right that making the keyclient notify systemd is probably the cleanest way to do this; I didn't really want to go through the effort of making that work, especially because it complicates the keyclient more, but I suppose that I probably should.

celskeggs:
I've made a bunch of changes and updated the description of this PR. Ready for review, though I'll be doing a bit more testing before I merge it.

building/build-debs/homeworld-admin-tools/src/deploy.py
```diff
@@ -4,13 +4,24 @@
...
+    spec = configuration.get_single_kube_spec(spec_name).encode()
+    for node in config.nodes:
+        if node.kind == "supervisor":
+            ssh.check_ssh(node, "mkdir", "-p", DEPLOYQUEUE)
+            ssh.upload_bytes(node, spec, "%s/%d.%s" % (DEPLOYQUEUE, int(time.time()), spec_name))
+            print("Uploaded spec to deployqueue.")
+    #with tempfile.TemporaryDirectory() as d:
+    #    specfile = os.path.join(d, "spec.yaml")
+    #    util.writefile(specfile, configuration.get_single_kube_spec(spec_name).encode())
+    #    access.call_kubectl(["apply", "-f", specfile], return_result=False)
```
ichung: commented-out code

PR #233, opened 2018-01-26
Repackage dns-addon containers

### LINKED ISSUE #32: Repackage the kubernetes dns-addon to use our own containers
We use the kubernetes DNS addon, as seen in //deployment/clustered/dns-addon.yml.

This is a service that runs on the cluster itself and helps with resolution of certain cluster-specific domains.

One general policy of the Homeworld project is that, to the greatest extent possible, we do *not* trust binaries built by upstream projects -- only source code. We especially do not trust binaries hosted by third parties. Unfortunately, the docker ecosystem is based on fetching containers from untrusted third parties, and the dns addon is no exception.

We need to repackage the relevant containers (kubedns, kube-dnsmasq, exechealthz) so that they are:

 * built entirely from source code
 * with upstream code vendored in our upstream vendoring repository
 * with code verified through signed upstream sources if at all possible (or through HTTPS connections from multiple users if not)
 * built as ACI containers instead of docker containers
 * built using our existing container build infrastructure
 * uploaded and stored in our container registry with existing upload infrastructure

Once we do this, the dns addon will become significantly more trustworthy.

### LINKED ISSUE #120: dns-addon shows abnormally high restart rates
We're up to 18 restarts within the first hour of DC5 having the dns-addon deployed. That's bad. What's going on?

celskeggs:
Fixes #32 but not #120.

Also updates dns-addon from 1.14.5 to 1.14.8.


building/build-acis/dnsmasq/dnsmasq.conf
```diff
@@ -0,0 +1 @@
+user=root
```
ichung: Why do we need dnsmasq to run as root?

celskeggs: It's inside a container. Running as root is a default.

celskeggs:
@ichung: any other concerns, or can I get you to approve this PR?

celskeggs:
Just rebasing.

Issue #182, opened 2017-12-28
Add flags for dry-run modes for spire seq commands


celskeggs:
This is because you might want to know the commands that are going to be executed, but not actually execute them.

Also, "don't print at all" should be an option. See comments on #178.

PR #237, opened 2018-01-27
Use argparse for argument parsing

### LINKED ISSUE #181: Support flags in spire
Currently, we only support command-line parameters, not flags. We should fix this.

krawthekrow:
Fixes #181 and #182. I had it no longer print the command sequence during the actual run, since it can be easily accessed with --dry-run anyway.

Also adds a manual command for checking supervisor certs (bug found during testing).

### DIFF
--- a/building/build-debs/homeworld-admin-tools/debian/changelog
+++ b/building/build-debs/homeworld-admin-tools/debian/changelog
@@ -1,3 +1,9 @@
+homeworld-admin-tools (0.1.41) stretch; urgency=medium
+
+  * Updated debian release
+
+ -- Mark Theng <mtheng@mit.edu>  Mon, 29 Jan 2018 15:52:39 -0500
+
 homeworld-admin-tools (0.1.40) stretch; urgency=medium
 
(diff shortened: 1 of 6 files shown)


andersk:
As a rule, I’d recommend avoiding argparse in favor of optparse, because the argparse developers [don’t understand how parsers are supposed to work](https://bugs.python.org/issue9334).

krawthekrow:
On the other hand, I'm not sure if we want to use optparse, given that it's deprecated. It also doesn't support sub-commands, but I can probably work around that. Anyway, our three obvious options are:

1. Use argparse.
2. Use optparse.
3. Parse command-line arguments ourselves.

Personally, I'm in favor of argparse. It's the easiest (if not simplest) option, and we wouldn't be affected by that bug anyway since we parse required arguments differently.

Caveat: I meant to test that kubectl with dashed arguments still works but haven't gotten to it yet. Though, it looks like it should work, based on what I read from the documentation.

celskeggs:
Rebase?

Issue #179, opened 2017-12-28
Non-descriptive error code from keyreq


krawthekrow:
Keyreq's error code does not allow us to differentiate between the connection being down, invalid client tickets, or a broken server keygateway.

celskeggs:
@krawthekrow: I don't remember the context for this; would you consider it sufficient to simply have keyreq print out more information on stderr?

krawthekrow:
Keyreq, on error, always exits with error code 1, often with a non-descriptive error message. The error messages are currently distinct enough to diagnose from Spire by text matching, but that's a rather ugly solution. It would be convenient (and more in line with other utilities, like SSH) if keyreq returns different exit codes for different classes of errors (and also have more descriptive error messages, if it is intended to be used outside Spire).

celskeggs:
It doesn't look possible to me to distinguish some of the above cases by the current output -- i.e. "server is down" versus "don't have tickets". But I can some of those along with adding more error codes.

elefthei:
Looks like similar to this one I've been working on: https://github.com/sipb/homeworld/issues/176
@krawthekrow Agree keyreq needs better error messages

celskeggs:
@krawthekrow: would you like to take this one? I worked on it previously (see the descriptive-error-codes branch), but never finished.

celskeggs:
Fixed by #261.

Issue #176, opened 2017-12-28
Better error messages when a request is made for a token for an invalid hostname


celskeggs:
This is the error that I got when I wrote `grilled-cheese` instead of `grilled-cheese.mit.edu`.

```
user@hyades-admin:~/dc6-proto1$ spire infra admit grilled-cheese
[keyreq] 2017/12/28 14:05:51.251465 keyreq.go:160: while unpacking json: unexpected end of JSON input ('[{"api":"bootstrap","body":"grilled-cheese"}]' -> '')
Traceback (most recent call last):
  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/bin/spire/__main__.py", line 35, in <module>
  File "/usr/bin/spire/command.py", line 76, in main_invoke
  File "/usr/bin/spire/command.py", line 27, in invoke
  File "/usr/bin/spire/command.py", line 27, in invoke
  File "/usr/bin/spire/command.py", line 66, in invoke
  File "/usr/bin/spire/infra.py", line 8, in infra_admit
  File "/usr/bin/spire/access.py", line 53, in call_keyreq
  File "/usr/lib/python3.5/subprocess.py", line 316, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.5/subprocess.py", line 398, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['keyreq', 'bootstrap-token', '/tmp/tmpy_fgcr8d/server.pem', 'egg-sandwich.mit.edu:20557', 'grilled-cheese']' returned non-zero exit status 1
```

elefthei:
I believe in this case the error should be reported in stderr by `keyreq` since the args passed are opaque to spire (doesn't know what's a hostname and what's a FQDN).

elefthei:
https://github.com/sipb/homeworld/pull/250

PR #249, opened 2018-02-03
Do a DNS lookup before calling request on keyreq, get better errors

elefthei:
Addresses https://github.com/sipb/homeworld/issues/176 
By doing a DNS lookup for the hostnames before trying to send a request. Maybe wrap the request code in that so we don't have to do it explicitly?

Will follow-up with more fixes for
https://github.com/sipb/homeworld/issues/179 if nobody claims it.

### DIFF
--- a/building/build-debs/homeworld-keysystem/src/keyreq/main/keyreq.go
+++ b/building/build-debs/homeworld-keysystem/src/keyreq/main/keyreq.go
@@ -3,6 +3,7 @@ package main
 import (
 	"log"
 	"os"
+	"net"
 	"keycommon/reqtarget"
 	"io/ioutil"
 	"keycommon/server"
@@ -37,6 +38,14 @@ func auth_kerberos(logger *log.Logger, authority_path string, keyserver_domain s
 		logger.Fatal("failed to check connection: ", err)
 	}
 	return ks, rt
+}
+
+func is_domain(host string) (bool) {
+        names, err := net.LookupHost(host)
+        if err != nil || len(names) == 0 {
+                return false;
+        }
+        return true;
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


elefthei:
Should the yaml config be passed as an argument on keyreq? It makes more
sense than hardcoding the path I think.

celskeggs:
I would recommend doing this validation in spire.

elefthei:
I'm not 100% sure that's the best thing to do, as you can see in `access.py` the faulty argument `grilled-cheese` is blindly passed to keyreq as part of `params`, without keyreq knowing weather it's a URL or something else.

```
def call_keyreq(keyreq_command, *params, collect=False):
    config = configuration.get_config()
    keyserver_domain = config.keyserver.hostname + "." + config.external_domain + ":20557"
    invoke_variant = subprocess.check_output if collect else subprocess.check_call

    with tempfile.TemporaryDirectory() as tdir:
        https_cert_path = os.path.join(tdir, "server.pem")
        util.writefile(https_cert_path, authority.get_pubkey_by_filename("./server.pem"))
        try:
            return invoke_variant(["keyreq", keyreq_command, https_cert_path, keyserver_domain] + list(params))
        except subprocess.CalledProcessError as e:
            if e.returncode == 1:
                fail_hint = "do you have valid kerberos tickets?\n" \
                    "or, your connection to the server might be faulty.\n" \
                    "or, the server's keygateway might be broken."
                command.fail("keyreq failed: %s" % e, fail_hint)
            raise e
```

So only `keyreq` would know how to make sense of these params. Unless you want to move out the target URL of `params` and make it an extra argument of `call_keyreq`. Let's just say that spire admit always has a target_URL argument and make the rest of its arguments part of `params`?

PR #250, opened 2018-02-07
Check configuration before making request on keyreq, get better errors

elefthei:
Something like this should be better for https://github.com/sipb/homeworld/issues/176 ?

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/configuration.py
+++ b/building/build-debs/homeworld-admin-tools/src/configuration.py
@@ -164,6 +164,18 @@ class Config:
                 return node
         command.fail("no such node: %s" % node_name)
 
+    def get_fqdn(self, name: str) -> str:
+        hostname = name
+        if name.endswith("." + self.external_domain):
+            # strip external domain
+            hostname = name[:-(len(self.external_domain) + 1)]
+        elif name.endswith("." + self.internal_domain):
+            # strip internal domain
+            hostname = name[:-(len(self.internal_domain) + 1)]
+        if not self.has_node(hostname):
+            command.fail("no such node: %s" % name)
+        return hostname + "." + self.external_domain
+
     @classmethod
     def load_from_string(cls, contents: bytes) -> "Config":
         return Config(yaml.safe_load(contents))
--- a/building/build-debs/homeworld-admin-tools/src/infra.py
+++ b/building/build-debs/homeworld-admin-tools/src/infra.py
@@ -5,7 +5,9 @@ import setup
 
 
 def infra_admit(server_principal: str) -> None:
-    token = access.call_keyreq("bootstrap-token", server_principal, collect=True)
+    config = configuration.get_config()
+    principal_hostname = config.get_fqdn(server_principal)
+    token = access.call_keyreq("bootstrap-token", principal_hostname, collect=True)
     print("Token granted for %s: '%s'" % (server_principal, token.decode().strip()))
 
 


building/build-debs/homeworld-admin-tools/src/infra.py
```diff
@@ -5,6 +5,9 @@
 
 
 def infra_admit(server_principal: str) -> None:
+    config = configuration.get_config()
+    if not config.has_node(server_principal):
+        command.fail("Server %s not found in project configuration" % server_principal)
```
celskeggs: Have you tested this? I can't remember exactly, but I seem to remember that `has_node` uses unqualified hostnames, whereas this method needs fully qualified hostnames.

elefthei: I think I need the real setup.yml to test it with real nodes, otherwise can I still only try `admit`

celskeggs: You should be able to access the real setup.yaml under https://github.mit.edu/sipb/hyades-cluster

celskeggs: Also, you'll need access to the cluster. Do you have a /root instance?

elefthei:
I do, my username slash root

celskeggs:
Access granted.

elefthei:
Changed the logic since it became kind of weird with the distinction between node_names and hostnames. Imo both should be accepted as argumemts, which is what the PR does now. It was tested on my local build.

building/build-debs/homeworld-admin-tools/src/infra.py
```diff
@@ -5,7 +5,9 @@
 
 
 def infra_admit(server_principal: str) -> None:
-    token = access.call_keyreq("bootstrap-token", server_principal, collect=True)
+    config = configuration.get_config()
+    principal_hostname = config.get_hostname(server_principal)
+    token = access.call_keyreq("bootstrap-token", principal_hostname, collect=True)
```
celskeggs: Good thinking. Coercing the name is definitely the correct solution to this problem.

building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -164,6 +164,27 @@ def get_node(self, node_name: str) -> Node:
...
         command.fail("no such node: %s" % node_name)
 
+    def has_hostname(self, hostname: str) -> bool:
+        if hostname.endswith(self.external_domain):
+            # strip external domain
+            hostname = hostname[:-(len(self.external_domain) + 1)]
+        elif hostname.endswith(self.internal_domain):
+            # strip internal domain
+            hostname = hostname[:-(len(self.internal_domain) + 1)]
+        return self.has_node(hostname)
```
celskeggs: It doesn't seem like this is used anywhere, so it should probably be removed, especially since it seems to duplicate code.

building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -164,6 +164,27 @@ def get_node(self, node_name: str) -> Node:
...
+
+    def get_hostname(self, name: str) -> str:
+        hostname = name
+        if name.endswith(self.external_domain):
+            # strip external domain
+            hostname = name[:-(len(self.external_domain) + 1)]
```
celskeggs: You're stripping out the dot in the domain name, but not checking that it exists. Given an `external_domain` of `mit.edu`, this could lead to a situation where something like `test.commit.edu` gets converted into `test.co`, which isn't good behavior.

building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -164,6 +164,27 @@ def get_node(self, node_name: str) -> Node:
...
+    def has_hostname(self, hostname: str) -> bool:
+        if hostname.endswith(self.external_domain):
+            # strip external domain
+            hostname = hostname[:-(len(self.external_domain) + 1)]
+        elif hostname.endswith(self.internal_domain):
+            # strip internal domain
+            hostname = hostname[:-(len(self.internal_domain) + 1)]
+        return self.has_node(hostname)
+
+    def get_hostname(self, name: str) -> str:
```
celskeggs: Nitpick: it might be worth making the point of this function clearer in the name. Maybe `lookup_hostname` or `validate_domain` or `convert_fqdn` or something along one of those lines?

building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -164,6 +164,27 @@ def get_node(self, node_name: str) -> Node:
...
+        hostname = name
+        if name.endswith(self.external_domain):
+            # strip external domain
+            hostname = name[:-(len(self.external_domain) + 1)]
+        elif name.endswith(self.internal_domain):
+            # strip internal domain
+            hostname = name[:-(len(self.internal_domain) + 1)]
```
celskeggs: Nitpick: the logic here might be clearer if the default case (`hostname = name`) were handled in an `else:` block.

building/build-debs/homeworld-admin-tools/src/configuration.py
```diff
@@ -164,6 +164,27 @@ def get_node(self, node_name: str) -> Node:
                 return node
```
celskeggs: A few meta notes:
 - Once you're done making changes, please squash your changes into a single commit.
 - Please fix the author so that GitHub recognizes this as your commit.
 - Please sign your commit with your PGP key. (Create a new key if you don't have one yet.)

building/build-debs/homeworld-admin-tools/src/configuration.py
elefthei: oops good catch

elefthei:
Latest comments by @celskeggs addressed, squashed and signed.

elefthei:
Pulled from master, should be up-to-date now

elefthei:
like this?

celskeggs:
No. That's rebasing master against your branch. You want to rebase your branch against master. You should only see one commit on github when you do this correctly.

PR #252, opened 2018-02-10
Upgrade linux, qemu, flannel and fix flannel connectivity bug

### LINKED ISSUE #95: Upgrade linux/qemu to more recent versions (for rkt kvm stage1)
Latest: 4.14.16, 2.11.0.

~This is intertwined with #91.~ Apparently upstream doesn't feel like using the latest versions of anything.

### LINKED ISSUE #247: Flannel connectivity seems to be broken in certain situations
This comes up occasionally, where a deployment will just happen to fail such that prometheus can't use the `eggs-benedict` (master01) node to connect to the flannel-monitor-collector service.

I suspect that this connectivity problem occurs exactly when a node has been assigned the 172.18.0.0/24 subnet (out of the 172.18.0.0/16 address space) and the connection is started from the node itself, rather than from a container running on the node.

We would only notice this when the node in question happens to be master01, because prometheus routes through master01 by default in the current configuration.

This needs to be checked by redeploying a cluster, and directly attempting to curl various flannel endpoints from each node. In the current cluster, direct access from the node to the cluster works on all nodes except master01, which is currently assigned 172.18.0.0/24. My hypothesis is that whichever node is assigned 172.18.0.0/24 will fail to access the rest of the flannel network.

### LINKED ISSUE #248: Upgrade flannel to latest upstream version: v0.10.0
Upstream: [v0.10.0](https://github.com/coreos/flannel/releases/tag/v0.10.0)

Upstream update checklist:
- [x] Ensure upstream version is latest.
- [x] Review changelogs for breaking changes.
- [x] Verify upstream release hash.
- [x] Add upgrade-flannel-0.10.0 branch with update.
- [x] Implement any repackaging changes.
- [x] Test cluster deployment with changes.
- [ ] Merge changes.

celskeggs:
Fixes #95. Fixes #248. Fixes #247.


celskeggs:
I've verified the upstream sources with a separate machine. Bypassing review for this PR, per zephyr conversation.

PR #254, opened 2018-02-11
Add system for redownloading files in the upstream repository

### LINKED ISSUE #230: Procedure for reproducing upstream tarballs should be documented
We should write down instructions on how to create the tarballs so that they can be reproduced independently of the homeworld-upstream repo.

celskeggs:
Fixes #230.

Also bumps the version of the debian ISO, because the old version is no longer kept in repositories, and rotates the PGP signing key, because I lost the key material due to a backup error. (Woops. I'll do a better job when we get closer to production.)

This PR is untested and should not be merged until testing is complete.


celskeggs:
Cluster deployed passed.

PR #261, opened 2018-02-18
More descriptive error messages from keyreq and spire

krawthekrow:
Alleviates #179. Currently does rather fragile error message parsing for the hint message, but I think it's fine since, if the messages change, the hints simply won't be displayed.

The alternative is to do the diagnosis within keyreq itself, but it feels like this might not be keyreq's responsibility.

### DIFF
--- a/building/build-debs/homeworld-admin-tools/src/access.py
+++ b/building/build-debs/homeworld-admin-tools/src/access.py
@@ -77,7 +77,7 @@ def diagnose_keyreq_error(errcode: int, err: str) -> Tuple[str, str]:
     return error_code_meaning, None
 
 
-def call_keyreq(keyreq_command, *params, collect=False):
+def call_keyreq(keyreq_command, *params):
     config = configuration.get_config()
     keyserver_domain = config.keyserver.hostname + "." + config.external_domain + ":20557"
 
--- a/building/build-debs/homeworld-admin-tools/src/access.py
+++ b/building/build-debs/homeworld-admin-tools/src/access.py
@@ -10,6 +10,7 @@ import base64
 import binascii
 import ssh
 import setup
+from typing import Tuple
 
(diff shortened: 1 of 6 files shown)


building/build-debs/homeworld-admin-tools/src/access.py
```diff
@@ -52,15 +64,32 @@ def call_keyreq(keyreq_command, *params, collect=False):
...
+                        fail_hint = "the server's keygateway might be broken."
+                    elif "gss_init_sec_context: No Kerberos credentials available" in knc_stderr or "gstd_error: gss_init_sec_context: Ticket expired" in knc_stderr:
+                        fail_hint = "do you have valid kerberos tickets?"
+                elif "empty response, likely because the server does not recognize your Kerberos identity" in err:
+                    fail_hint = "your kerberos tickets might be for the wrong instance."
+            fail_message = "keyreq failed with error code %d: %s" % (keyreq_sp.returncode, error_code_meaning)
+            if fail_hint is None:
+                command.fail(fail_message)
+            else:
+                command.fail(fail_message, fail_hint)
```
celskeggs: This entire block of added code is fairly messy, and hard to decipher. Can you try to find some way to clean it up and make it more concise?

building/build-debs/homeworld-keysystem/debian/changelog
```diff
@@ -1,3 +1,33 @@
...
+  * Update debian version
+
+ -- Cel Skeggs <cela@mit.edu>  Fri, 02 Feb 2018 15:38:58 -0500
+
+homeworld-keysystem (0.1.17) stretch; urgency=medium
+
+  * Update debian version
+
+ -- Cel Skeggs <cela@mit.edu>  Fri, 02 Feb 2018 15:15:39 -0500
+
```
celskeggs: It's probably worth collapsing these different keysystem version increments into one.

krawthekrow:
Also, currently, I'm just ignoring the "collect" parameter. Could I just remove it from all calls of call_keyreq?

PR #275, opened 2018-03-25
Switch to full-project build isolation from per-component build isolation

### LINKED ISSUE #215: Add a rkt build helper so that we can separate our deployable rkt from our development rkt
Currently, you have to build the on-cluster version, and install it locally, and then make sure to switch it to non-KVM mode every time you use it.

We should just build a version for local use.

### LINKED ISSUE #218: Switch from per-component build isolation to full-project build isolation
Currently, individual pieces of the build will use various systems -- schroot, rkt -- to try to isolate themselves from variations in the host system.

We should have a build container (or just plain chroot) that we use for everything. It would have the normal workspace bind-mounted, and would be a consistent environment for *everything*. The only things that would be built outside of this container are the container itself and a local copy of rkt to launch it.

This would simplify and unify our build isolation tooling, and help us move towards a less complicated build system overall.

celskeggs:
Fixes #218. Fixes #215. The idea here is that we only need *one* chroot, rather than a large number of them and a build container.

I wasn't able to figure out a good way to use existing chroot management tools in a way that made this easier. Perhaps I should try harder.

Even though this isn't a net code reduction, it should open the doors for further code reduction.

This hasn't been fully tested.


building/build-acis/common/container-build-helpers.sh
```diff
@@ -53,6 +57,7 @@ function allocate_tempdir() {
 		echo "Could not create temporary directory." 1>&2
 		exit 1
 	fi
+	B="${TMPBUILDDIR}"
```
krawthekrow: Why reassign TMPBUILDDIR? Presumably B would always be the same as TMPBUILDDIR, since the if statement exits. Also, I have reservations about the variable name being non-descriptive and hard to grep, but I can see the case for it being short here, so I won't let this block it.

create-chroot.sh
```diff
@@ -0,0 +1,32 @@
...
+if [ -e "${HOMEWORLD_CHROOT}" ]
+then
+    echo "chroot already exists" 1>&2
+    exit 1
+fi
+
+mkdir "${HOMEWORLD_CHROOT}"
+sudo debootstrap --include="$(cat chroot-packages.list | grep -vE '^#' | tr '\n ' ',,' | sed 's/,$//')" stretch "${HOMEWORLD_CHROOT}" http://debian.csail.mit.edu/debian/
+sudo ln -sT "homeworld/building" "${HOMEWORLD_CHROOT}/h"
+sudo chroot "${HOMEWORLD_CHROOT}" useradd -m -u "$UID" "$USER" -s "/bin/bash"
```
krawthekrow: Why specify the uid of the new user?

create-chroot.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [ -e "${HOMEWORLD_CHROOT}" ]
+then
+    echo "chroot already exists" 1>&2
+    exit 1
+fi
+
+mkdir "${HOMEWORLD_CHROOT}"
+sudo debootstrap --include="$(cat chroot-packages.list | grep -vE '^#' | tr '\n ' ',,' | sed 's/,$//')" stretch "${HOMEWORLD_CHROOT}" http://debian.csail.mit.edu/debian/
+sudo ln -sT "homeworld/building" "${HOMEWORLD_CHROOT}/h"
```
krawthekrow: Shouldn't the directory you're linking be building (instead of homeworld/building)? Also shouldn't all the chroot scripts be logically under building instead of the project root?

building/build-acis/common/container-build-helpers.sh
celskeggs: Line 54 is the only place that TMPBUILDDIR is assigned.

B is a shorthand. You can grep for it as either `${B}` or `$B[^A-Z]`. I also have similar reservations, but it seemed worthwhile.

create-chroot.sh
celskeggs: The idea was to make sure it matches the actual user, because the filesystem is shared. I feel like there might be another solution to this, but I can't remember what it is.

create-chroot.sh
celskeggs: I didn't want to have it directly under the project directory, because I didn't want a stray 'rm -rf' to erase a mountpoint on the real filesystem. This concern might be mitigated now that we're using nspawn-based isolation.

PR #284, opened 2018-05-19
Migrate to cleaner build system

celskeggs:
Previously, our build code was mostly copy-paste spaghetti.

Now it's slightly less spaghetti and significantly less copy-paste.

This PR is now ready for review.


celskeggs:
To review this, the best process is probably:

 - Glance at some glass.yaml files to understand the build format.
 - Review the final version of the glass source code.
 - Review each individual build script migration commit-by-commit.

celskeggs:
Note: I should confirm that these build changes didn't break the cluster before merging.

building/glass/inner.py
```diff
@@ -0,0 +1,612 @@
...
+        self.steps = []
+        self.branch = branch
+
+    def add_step(self, step_func):
+        self.steps.append(step_func)
+
+    # returns a function that can be called with the build tree
+    def add_parse(self, step_yaml: yamlobj.YamlObject):
+        step_type = step_yaml["type"].string()
+        if step_type == "upstream-extract":
```
ichung: abstract the branches of this giant if/else

ichung:
Further comments:
- "yamlobj" isn't actually about YAML (besides the load function); it's a small but generic typechecker for python objects.
- Instead of lazily typecheck objects as they're consumed, I suggest to validate the entire loaded object once with a schema (say, in the style of https://github.com/Grokzen/pykwalify), which will significantly simplify interaction with the loaded objects.
- inner.py and acbuild.py are both rather convoluted in the same way.  I'm not sure what the benefit of all the thunking and prepared-execution is over just performing actions directly.

building/glass/acbuild.py
```diff
@@ -0,0 +1,115 @@
...
+        self._exec("mount", "add", "--", mountname, mountpoint)
+
+    def label_add(self, key, value):
+        self._exec("label", "add", "--", key, value)
+
+    def port_add(self, portname, proto, portnum):
+        self._exec("port", "add", "--", portname, proto, portnum)
+
+    def write(self, output_file, time_unix: int, tempdir: str):
+        with tempfile.TemporaryDirectory(prefix="acbuild-out-", dir=tempdir) as interstitial:
```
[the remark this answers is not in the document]
celskeggs: Good point.

building/glass/acbuild.py
```diff
@@ -0,0 +1,116 @@
...
+                subprocess.check_call(["tar", "-C", extractdir, "-cf", post_repack, "."])
+            with open(output_file, "wb") as f:
+                subprocess.check_call(["gzip", "-n", "-c", post_repack], stdout=f)
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._exec("end")
+        self._tempdir_obj.cleanup()
+
+
+class Prepared:
```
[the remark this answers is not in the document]
celskeggs: Hmm, probably.

building/glass/aptbranch.py
```diff
@@ -0,0 +1,73 @@
+"This module handles resolving which apt branches exist, and what their corresponding signing keys are."
+
+import os
+import re
+import subprocess
+
+SETUP_DIR = "/h/setup-apt-branch/"
```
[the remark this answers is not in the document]
celskeggs: A fixed, globally-known path in the chroot, which points to homeworld/building.

building/glass/aptbranch.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def get_env_branch():
+    """Get the apt branch named in the environment variable ${HOMEWORLD_APT_BRANCH}.
+    If should_error_on_nonexistent is true or unspecified, throw an exception if the environment variable isn't found.
```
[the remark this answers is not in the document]
celskeggs: Woops, I'll remove this outdated documentation.

building/glass/aptbranch.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if not re.match("^[0-9a-zA-Z_.-]+/[0-9a-zA-Z_.-]+$", branch):
+        print("apt branch should be of the form <username>/<branch>")
+        print("allowed characters: [0-9a-zA-Z_.-]")
+        raise Exception("invalid apt branch: %s" % branch)
+
+    return branch
+
+
+def load_key_value_pairs(filename):
```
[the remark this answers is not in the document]
celskeggs: This is a migration of existing code for what that file looks like. It was in that format for easier parsing from bash. We still have bash scripts parsing that file, so I don't want to change the format just yet.

building/glass/glass.py
```diff
@@ -0,0 +1,50 @@
...
+import aptbranch
+import inner
+
+if not os.path.exists("/h/"):
+    raise Exception("not currently running within the homeworld build chroot")
+
+parser = argparse.ArgumentParser(description="perform a glass build")
+parser.add_argument("-b", "--branch", default=aptbranch.get_env_branch(),
+                    help="the apt branch to build this package for")
+parser.add_argument("-d", "--debug", type=bool, nargs="?", const=True, default=False,
```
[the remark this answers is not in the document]
celskeggs: Doesn't that not let us do '--debug=false'? (Which might be useful if a caller is composing a complicated command line.)

building/glass/inner.py
```diff
@@ -0,0 +1,612 @@
...
+            log("debug", "running with debug mode on")
+
+        self.ensure_create_bin_path(branch)
+        self.ensure_package_name_matches()
+
+        os.unsetenv("GOROOT")
+
+        tempdir = None  # use default
+
+        if self.glassfile["control"].has_key("use-tmpfs") and not self.glassfile["control"]["use-tmpfs"].boolean():
```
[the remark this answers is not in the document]
celskeggs: We have the get method already. But yes, we should use it here.

building/glass/inner.py
celskeggs: Good point. Might not get to that right now.

celskeggs:
The benefit of thunking in the general case is that we can do syntax-checking of the entire glass.yaml before starting to execute, which saves significant iteration time.

ichung:
Checking the yaml file against a schema could address that issue.

celskeggs:
Yeah.

building/glass/inner.py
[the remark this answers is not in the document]
ichung: I think the existing get method would need to be modified a little in order for this to interact properly with typechecking.

celskeggs:
@ichung: re-review? I think I've addressed your comments.

building/glass/debuild.py
```diff
@@ -0,0 +1,83 @@
...
+
+        write_file(debdir, "%s.dirs" % name,
+                   "\n".join(os.path.relpath(tree, dirpath) for dirpath, _, _ in os.walk(tree)))
+
+        validate_install_script_names(install_scripts.keys())
+        for iname, ipath in install_scripts.items():
+            with open(ipath, "r") as f:
+                write_file(debdir, iname, f.read())
+
+        subprocess.check_call(["debuild", "-us", "-uc", "-b"], cwd=builddir, stdout=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
celskeggs: Switching to these causes the debuild command to fail. I have no idea why, but this isn't worth my debugging time.

building/glass/acbuild.py
```diff
@@ -0,0 +1,78 @@
...
+        self._exec("port", "add", "--", portname, proto, portnum)
+
+    def write(self, output_file, time_unix: int, tempdir: str):
+        with tempfile.TemporaryDirectory(prefix="acbuild-out-", dir=tempdir) as interstitial:
+            pre_repack = os.path.join(interstitial, "pre-repack.aci")
+            self._exec("write", pre_repack)
+
+            # based on https://bitbucket.org/ikdc/tarclean/
+            with tarfile.open(pre_repack, mode='r') as t_in:
+                with tarfile.open(output_file, mode='w|gz') as t_out:
```
andersk: If an exception is thrown in this block, the partial output file will still be around, so future runs of glass will assume that the build succeeded. You should write to another temporary file and rename it into place at the end.

(In my test build, a `UnicodeDecodeError` was thrown in this block. I’m still trying to figure out why.)

celskeggs:
@ichung: do you have any ideas on the unicode error andersk reported, because the relevant code is derived from your tarclean stuff?

building/glass/acbuild.py
ichung: Regarding the encoding error: we didn't specify an encoding, so it ought to default to the filesystem encoding (tarfile.py:147), which is usually not ascii.  @andersk Could you check what `sys.getfilesystemencoding()` returns here?  Also, if you produce the tarfile in question somewhere then I will try and debug this one.

andersk: `sys.getfilesystemencoding()` returns `'ascii'`, since `debootstrap` doesn’t configure any locales. Did something different happen for you?

(This is maybe one of the problems with doing builds inside a mutable chroot instead of an sbuild-style ephemeral chroot—it’s easy to miss hidden dependencies that you configured manually or as a side effect of a different build.)

ichung:
I didn't run it yet, but cela can answer that question.  Probably we should just specify a more reasonable encoding.On Jun 5, 2018 03:14, Anders Kaseorg <notifications@github.com> wrote:@andersk commented on this pull request.



In building/glass/acbuild.py:
> +        self._exec("mount", "add", "--", mountname, mountpoint)
+
+    def label_add(self, key, value):
+        self._exec("label", "add", "--", key, value)
+
+    def port_add(self, portname, proto, portnum):
+        self._exec("port", "add", "--", portname, proto, portnum)
+
+    def write(self, output_file, time_unix: int, tempdir: str):
+        with tempfile.TemporaryDirectory(prefix="acbuild-out-", dir=tempdir) as interstitial:
+            pre_repack = os.path.join(interstitial, "pre-repack.aci")
+            self._exec("write", pre_repack)
+
+            # based on https://bitbucket.org/ikdc/tarclean/
+            with tarfile.open(pre_repack, mode='r') as t_in:
+                with tarfile.open(output_file, mode='w|gz') as t_out:

sys.getfilesystemencoding() returns 'ascii', since debootstrap doesn’t configure any locales. Did something different happen for you?
(This is maybe one of the problems with doing builds inside a mutable chroot instead of an sbuild-style ephemeral chroot—it’s easy to miss hidden dependencies that you configured manually or as a side effect of a different build.)

andersk:
After doing

```diff
--- a/building/glass/acbuild.py
+++ b/building/glass/acbuild.py
@@ -68,6 +68,6 @@ class Build:
             # based on https://bitbucket.org/ikdc/tarclean/
-            with tarfile.open(pre_repack, mode='r') as t_in:
+            with tarfile.open(pre_repack, mode='r', encoding='utf-8') as t_in:
                 if os.path.exists(output_file + ".tmp"):
                     os.unlink(output_file + ".tmp")
-                with tarfile.open(output_file + ".tmp", mode='w|gz') as t_out:
+                with tarfile.open(output_file + ".tmp", mode='w|gz', encoding='utf-8') as t_out:
                     for entry in sorted(t_in, key=lambda e: e.path):
```

the Debian chroots build, though they take _forever_ because `tarfile` is slow. But then I get

```
      [glass] beginning build of package dnsmasq-2.78-1-linux-amd64.aci for aptdomain/aptbranch
        [aci] unpacking container rootfs from debian-micro version stretch.20180529T215324Z to output:.
tar: ./rootfs: Not found in archive
tar: Exiting with failure status due to previous errors
Traceback (most recent call last):
  File "../glass/glass.py", line 52, in <module>
    p.run(args.branch, debug=args.debug)
  File "/homeworld/building/glass/project.py", line 127, in run
    project.run(branch)
  File "/homeworld/building/glass/project.py", line 172, in run
    raise e
  File "/homeworld/building/glass/project.py", line 166, in run
    actions.invoke_action(ctx, step)
  File "/homeworld/building/glass/actions.py", line 21, in invoke_action
    getattr(actions, "perform_%s" % step_type.replace("-", "_"))(ctx, **step_object)
  File "/homeworld/building/glass/actions.py", line 265, in perform_aci_unpack
    subprocess.check_call(["tar", "-C", targetdir, "-xf", aci, "--strip-components=2", "./rootfs"])
  File "/usr/lib/python3.5/subprocess.py", line 271, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['tar', '-C', '/tmp/glass-huyzocp9/tree/', '-xf', '/h/binaries/aptdomain/aptbranch/debian-micro-stretch.20180529T215324Z-linux-amd64.aci', '--strip-components=2', './rootfs']' returned non-zero exit status 2
```

because the archive contains `rootfs` instead of `./rootfs`.

ichung:
How slow?  If we're starting with compressed archives then tarfile will decompress from the beginning for each seek, so it is much faster to decompress entirely first and pass the decompressed archive to tarfile.  No idea about the path issue.  Though maybe it is not worth trying to get tarfile working and we should just go back to the version of this code which extracts and repacks the files.On Jun 5, 2018 05:35, Anders Kaseorg <notifications@github.com> wrote:After doing
--- a/building/glass/acbuild.py
+++ b/building/glass/acbuild.py
@@ -68,6 +68,6 @@ class Build:
             # based on https://bitbucket.org/ikdc/tarclean/
-            with tarfile.open(pre_repack, mode='r') as t_in:
+            with tarfile.open(pre_repack, mode='r', encoding='utf-8') as t_in:
                 if os.path.exists(output_file + ".tmp"):
                     os.unlink(output_file + ".tmp")
-                with tarfile.open(output_file + ".tmp", mode='w|gz') as t_out:
+                with tarfile.open(output_file + ".tmp", mode='w|gz', encoding='utf-8') as t_out:
                     for entry in sorted(t_in, key=lambda e: e.path):
the Debian chroots build, though they take forever because tarfile is slow. But then I get
      [glass] beginning build of package dnsmasq-2.78-1-linux-amd64.aci for aptdomain/aptbranch
        [aci] unpacking container rootfs from debian-micro version stretch.20180529T215324Z to output:.
tar: ./rootfs: Not found in archive
tar: Exiting with failure status due to previous errors
Traceback (most recent call last):
  File "../glass/glass.py", line 52, in <module>
    p.run(args.branch, debug=args.debug)
  File "/homeworld/building/glass/project.py", line 127, in run
    project.run(branch)
  File "/homeworld/building/glass/project.py", line 172, in run
    raise e
  File "/homeworld/building/glass/project.py", line 166, in run
    actions.invoke_action(ctx, step)
  File "/homeworld/building/glass/actions.py", line 21, in invoke_action
    getattr(actions, "perform_%s" % step_type.replace("-", "_"))(ctx, **step_object)
  File "/homeworld/building/glass/actions.py", line 265, in perform_aci_unpack
    subprocess.check_call(["tar", "-C", targetdir, "-xf", aci, "--strip-components=2", "./rootfs"])
  File "/usr/lib/python3.5/subprocess.py", line 271, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['tar', '-C', '/tmp/glass-huyzocp9/tree/', '-xf', '/h/binaries/aptdomain/aptbranch/debian-micro-stretch.20180529T215324Z-linux-amd64.aci', '--strip-components=2', './rootfs']' returned non-zero exit status 2

because the archive contains rootfs instead of ./rootfs.

celskeggs:
re: ascii: actually, I expect that this would also fail under my environment -- I only tested the new repacking with the smaller images, not the larger ones.

re: ./rootfs: woops

Honestly, I don't think this code's original goal (reproducible builds) is worth half-heartedly pursuing -- either it's all the way there, or it's useless. As such, I'm probably just going to delete the repacking code. It's not worth the debugging time.

celskeggs:
@ichung: I'm testing this now; would you like to re-review?

ichung:
Should I review now or wait until the build passes?

LICENSE
```diff
@@ -1,6 +1,14 @@
...
 
-Copyright (c) 2017 Cel Skeggs
+Copyright (c) 2017-2018 Cel Skeggs
+Copyright (c) 2017-2018 Mark Theng
+Copyright (c) 2017-2018 ikdc
+Copyright (c) 2018 Eleftherios Lef Ioannidis
+Copyright (c) 2018 David Sargent
+Copyright (c) 2017 Anders Kaseorg
+Copyright (c) 2017 Matthew Ince
+Copyright (c) 2017 Sam Dukhovni
```
ichung: Faye Dukhovni

building/glass/acbuild.py
```diff
@@ -0,0 +1,64 @@
+import os
+import subprocess
+import tarfile
```
ichung: unused `tarfile` import

enter-chroot.sh
```diff
@@ -9,4 +9,17 @@ then
...
+ORIG_REL="$(realpath --relative-to "$(pwd)/building" "${ORIG_PWD}")"
+if [[ "${ORIG_REL}" =~ \.\. ]]
+then
+	ORIG_REL=""
+fi
+if [ -e "$HOME/.gnupg/pubring.kbx" ]
+then
+	cp "$HOME/.gnupg/pubring.kbx" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/pubring.kbx"
+	cp "$HOME/.gnupg/trustdb.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/trustdb.gpg"
+	cp -R "$HOME/.gnupg/private-keys-v1.d/"* "$HOMEWORLD_CHROOT/home/$USER/.gnupg/private-keys-v1.d"
```
ichung: I'm concerned about copying the user's private keys into the build chroot.  Why is this necessary?

celskeggs: It's necessary in order to sign packages within the chroot, which is currently how the upload process works.

celskeggs:
I will be redoing all of these commits, in an attempt to clean up this PR and make it slightly easier to merge.

celskeggs:
I've restructured all of the commits, to better separate out different changes. Could be better, but probably good enough. It should now be possible to review this part-by-part, so I'll be splitting out some partial shorter pull requests that should be easier to review.

enter-chroot.sh
ichung: I think this could be done in a better way.
- I don't think `private-keys-v1.d` is a good interface to use; we should use `gpg --export-secret-key` instead.
- Along the same lines, we should move only the key we need for signing.

ichung:
Should I review all this code, or wait for you to further split off parts?

enter-chroot.sh
celskeggs: We don't know what key we'll need for signing at the point where we enter the chroot.

celskeggs:
I'll accept reviews in both places. Do whatever seems like the best use of your time.

enter-chroot.sh
ichung: How does the upload process determine which key to use?

celskeggs: It's based on the branch specified by the user, which is usually only specified once they're already within the chroot.

ichung: Can we at least take only the ones listed in `(global-)?signing-keys`?

celskeggs: We could try.

In any case, I tried this, and it turns out that then the user has to manually decrypt the relevant pgp keys to insert them into the chroot, and that's not especially fun. It'll require a lot of complicated synchronization code to do this, and since you already LGTM'd this part of the code in the other PR, I'm separating this out into a separate bug, #287, to be acted on later.

celskeggs:
@ichung: this will be difficult to split up into multiple commits, because the migration to the new build system needs to happen all at once -- so just review this PR.

celskeggs:
I successfully deployed a complete cluster using this version. It's now entirely in the code review court.

celskeggs:
I've pushed four additional commits from another branch, which should cover the doc fixes. Is that sufficient?

PR #285, opened 2018-06-09
Improve chroot system in preparation for glass build upgrade

celskeggs:
This is a smaller part of #284, which is very large.

This small set of changes should be easy to review.

### DIFF
--- a/enter-chroot.sh
+++ b/enter-chroot.sh
@@ -22,4 +22,4 @@ then
 	cp "$HOME/.gnupg/trustdb.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/trustdb.gpg"
 	cp -R "$HOME/.gnupg/private-keys-v1.d/"* "$HOMEWORLD_CHROOT/home/$USER/.gnupg/private-keys-v1.d"
 fi
-sudo systemd-nspawn -M homeworld --bind $(pwd):/homeworld:norbind -u "$USER" -a -D "$HOMEWORLD_CHROOT" bash -c "cd /h/${ORIG_REL} && gpg-agent --daemon --keep-tty && exec bash"
+sudo systemd-nspawn --bind $(pwd):/homeworld:norbind -u "$USER" -a -D "$HOMEWORLD_CHROOT" bash -c "cd /h/${ORIG_REL} && gpg-agent --daemon --keep-tty && exec bash"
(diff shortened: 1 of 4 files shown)


celskeggs:
Added one small unrelated commit.

enter-chroot.sh
```diff
@@ -9,4 +9,17 @@ then
     exit 1
 fi
 
-sudo systemd-nspawn -M homeworld --bind $(pwd):/homeworld:norbind -u "$USER" -a -D "$HOMEWORLD_CHROOT" bash -c "cd /h/ && exec bash"
+ORIG_PWD="$(pwd)"
+cd "$(dirname "$0")"
```
krawthekrow: It looks like this puts you in a different directory from where you started after the script exits. Is this intended?

enter-chroot.sh
```diff
@@ -9,4 +9,17 @@ then
...
+ORIG_REL="$(realpath --relative-to "$(pwd)/building" "${ORIG_PWD}")"
+if [[ "${ORIG_REL}" =~ \.\. ]]
```
krawthekrow: This might not be an edge case you care about, but I think this won't work if you're in a directory whose name contains "..".

enter-chroot.sh
celskeggs: Huh? The current working directory of the bash subprocess is different from the cwd of the shell.

enter-chroot.sh
celskeggs: I don't think I care about this edge case.

ichung: What is the purpose of this code?  If I understand, if you run the script from a subdirectory of `building` then it spanws the new shell in the corresponding subdirectory of the chroot.  While I understand that it might be useful to control the subdirectory of the chroot used, why is using the caller's pwd a good way to do so?

ichung: Also, why is this code only present in the systemd-nspawn variant?

celskeggs: It's a good way to do so because, if you have a terminal open to a subdirectory and want to run a build, you can easily launch a shell that points to the same context you've already browsed to.

This is a human-oriented feature, because it's annoying to cd into a directory, launch a chroot, and then cd into the same directory *again*. It smooths over the interface between the chroot and the outside world.

It is only present in the systemd-nspawn variant because ubuntu 14.04 cannot run `realpath --relative-to`, which makes it much harder to implement, and it's a human-oriented feature, so less useful for automated environments.

celskeggs:
For the sake of not slowing things down, I'm treating one approval as enough, and merging this now. Any remaining problems can be filed as issues.

PR #288, opened 2018-06-14
glass: add new build tool

celskeggs:
Please refer to the git commit messages for additional commentary on these changes.

This is part of the larger #284.


building/glass/aptbranch.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def get_signing_key(branch):
+    "Get the signing key ID for a specific apt branch. Throw an exception if not found, or if the key doesn't exist in the gpg keyring."
+
+    key_id = get_signing_key_unchecked(branch)
+    if subprocess.call(["gpg", "--list-keys", key_id], stdout=subprocess.DEVNULL) != 0:
+        if branch == "root/master":
+            print("If you're basing this build off the master branch, import its signing key with")
+            print('gpg --import \"$(readlink -f "${APT_SETUP_DIR}/../upload-debs/default-repo-signing-key.gpg")\"')
```
ichung: nit: the backslashes serve no purpose except to confuse me for 2 minutes :)

building/glass/debclean.py
```diff
@@ -0,0 +1,49 @@
+import os
+import shutil
+
+
+def clean_paths(rootfs, *paths):
```
ichung: This module seems like a good way to accidentally wipe a developer's filesystem.  I wonder if there's a good way to check so that it fails if not called inside the chroot?

celskeggs:
Fixed just the backslashes.

building/glass/debclean.py
ichung: I misread the surrounding code.  Nonetheless this would still be bad if called with the wrong `rootfs` argument, no?

celskeggs: Yes... but so would a lot of things? Like shutil.rmtree, for example.

We have two layers of protection: 1. the build chroot and 2. the fact that this is embedded in a larger program that can ensure it's used correctly.

Both of these would have to fail for it to cause serious damage to a developer's system... and that's why we have backups.

PR #291, opened 2018-06-28
Add image-based deployment system

### LINKED ISSUE #86: Switch from a preseed-based system to an image-based system
We should be preparing static images of the debian installs that we want, and flashing those onto cluster machines, instead of installing a complete debian system each time.

celskeggs:
Fixes #86.

THIS IS NOT READY FOR MERGE.

This is a prototype of a possible change to our deployment process, where we pre-generate images, and then deploy them onto actual machines mostly just by writing directly to disk.

This has been tested and works, but:

 - Some parts of this system are hardcoded in ways that are not acceptable for a final version.
 - This appears to slow down deployments from ~17 minutes to ~18 minutes, and it is not immediately clear why. If anything, this should speed up deployments.
 - It is not yet clear whether this is a worthwhile change overall; it is possible that this change will lead to unnecessary maintenance overhead, and should be scrapped on that basis.


celskeggs:
As it turns out, I'm still having trouble getting this change to work out, because something I did makes the entire bring-up process after installation finishes take significantly longer, which cancels out the gains from the faster install.

Having spent far too much time trying to make this work out, I'm putting this change on the shelf, and will come back to it once I've done more work to streamline deployment in other ways, and maybe also have better tools for understanding the performance properties of the deployment process.

PR #292, opened 2018-07-06
Allow us to configure glass to upload via rsync

krawthekrow:
~First step towards local deploy. I haven't tested this through a full deploy yet, but I wanted to get some feedback on my approach. It currently works up to installing homeworld-admin-tools on the deployment VM.~

This adds rsync as an alternative option for upload, so that developers can deploy without GCS. Since rsync requires a somewhat larger amount of configuration that would be the same across multiple apt branches on the same upload target, this is implemented so that the same upload target configuration can be reused across branches easily.

This is tested via #321 since it was necessary for deployment to work with my own repository signing key.

There isn't an issue assigned for this.

---

Checklist:

### DIFF
--- a/building/apt-branch-config/branches.yaml.example
+++ b/building/apt-branch-config/branches.yaml.example
@@ -1,6 +1,11 @@
 branches:
   - name: root/master
     signing-key: d167aca2e5fcf374
-  - name: <name>/<subbranch>
+  - name: <host>/<location>/<of>/<repository>
     signing-key: <signing key>
-    apt-url-prefix: homeworld.deploy
+    upload:
+      method: rsync
+	  # these will be passed to rsync as
+	  # <username>@<host>:<dir>/<location>/<of>/<repository>
+	  user: <username>
+      dir: /var/www/html
--- a/building/apt-branch-config/branches.yaml.example
+++ b/building/apt-branch-config/branches.yaml.example
@@ -1,11 +1,14 @@
+upload-targets:
+  - name: <upload target>
+    method: rsync
+    # these will be passed to rsync as
+	# <username>@<host>:<dir>/<location>/<of>/<repository>
+	user: <username>
+    dir: /var/www/html
+
 branches:
   - name: root/master
     signing-key: d167aca2e5fcf374
   - name: <host>/<location>/<of>/<repository>
     signing-key: <signing key>
-    upload:
-      method: rsync
-	  # these will be passed to rsync as
-	  # <username>@<host>:<dir>/<location>/<of>/<repository>
-	  user: <username>
-      dir: /var/www/html
+    upload-target: <upload target>
--- a/building/glass/aptbranch.py
+++ b/building/glass/aptbranch.py
@@ -55,11 +55,20 @@ def select_branch_config(branches_config: list, branch: str):
     raise Exception("no config found for %s in %s" % (branch, CONFIG_PATH))
 
 
+def select_upload_config(upload_targets: list, name: str):
+    for upload_target in upload_targets:
+        if upload_target["name"] == name:
+            return upload_target
+    raise Exception("no config found for %s in upload-targets" % name)
+
+
 class Config:
     def __init__(self, branch: str):
         if not os.path.exists(CONFIG_PATH):
             raise Exception("cannot find branches config at %s, use %s.example to create one" % (CONFIG_PATH, CONFIG_PATH))
-        branches_config = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)["branches"]
+        configs = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)
+        branches_config = configs["branches"]
+        self.upload_targets = configs.get("upload-targets", [])
         self.config = select_branch_config(branches_config, branch)
 
         check_signing_key(self.signing_key)
@@ -74,6 +83,7 @@ class Config:
     
     @property
     def upload_config(self):
-        if "upload" not in self.config:
-            raise Exception("no upload config found in branch %s" % self.name)
-        return self.config["upload"]
+        if "upload-target" not in self.config:
+            raise Exception("no upload target found for branch %s" % self.name)
+        upload_target_name = self.config["upload-target"]
+        return select_upload_config(self.upload_targets, upload_target_name)
--- a/building/glass/aptbranch.py
+++ b/building/glass/aptbranch.py
@@ -6,6 +6,9 @@ import subprocess
 
 import validate
 
+
+APT_BRANCH_REGEX = "^[0-9a-zA-Z_.-][0-9a-zA-Z_./-]+[0-9a-zA-Z_.-]$"
+
 CONFIG_PATH = "/h/apt-branch-config/branches.yaml"
 CONFIG_SCHEMA_NAME = "branches-schema.yaml"
 
@@ -18,9 +21,10 @@ def get_env_branch():
     if not branch:
         return None
 
-    if not re.match("^[0-9a-zA-Z_.-]+/[0-9a-zA-Z_.-]+$", branch):
-        print("apt branch should be of the form <username>/<branch>")
-        print("allowed characters: [0-9a-zA-Z_.-]")
+    if not re.match(APT_BRANCH_REGEX, branch):
+        print("apt branch should be a url path")
+        print("e.g. homeworld.apt/branch/subbranch")
+        print("allowed characters: [0-9a-zA-Z_./-]")
         raise Exception("invalid apt branch: %s" % branch)
 
     return branch
@@ -67,10 +71,9 @@ class Config:
     @property
     def signing_key(self) -> str:
         return self.config["signing-key"]
-
+    
     @property
-    def apt_url(self):
-        url_prefix = self.config.get("apt-url-prefix", "")
-        if url_prefix and not url_prefix.endswith("/"):
-            url_prefix += "/"
-        return url_prefix + self.name
+    def upload_config(self):
+        if "upload" not in self.config:
+            raise Exception("no upload config found in branch %s" % self.name)
+        return self.config["upload"]
--- a/building/glass/branches-schema.yaml
+++ b/building/glass/branches-schema.yaml
@@ -9,12 +9,27 @@ properties:
       properties:
         name:
           type: string
-          pattern: "^[0-9a-zA-Z_.-]+/[0-9a-zA-Z_.-]+$"
+          pattern: "^[0-9a-zA-Z_.-][0-9a-zA-Z_./-]+[0-9a-zA-Z_.-]$"
         signing-key:
           type: string
           pattern: "^[a-fA-F0-9]+$"
-        apt-url-prefix:
-          type: string
+        upload:
+          type: object
+          oneOf:
+          - properties:
+              method:
+                enum: ["google-cloud-storage"]
+            required: ["method"]
+            additionalProperties: false
+          - properties:
+              method:
+                enum: ["rsync"]
+              user:
+                type: string
+              dir:
+                type: string
+            required: ["method", "user", "dir"]
+            additionalProperties: false
       required: ["name", "signing-key"]
       additionalProperties: false
 required: ["branches"]
--- a/building/glass/branches-schema.yaml
+++ b/building/glass/branches-schema.yaml
@@ -2,6 +2,29 @@ $schema: "http://json-schema.org/draft-07/schema#"
 title: BranchesConfig
 type: object
 properties:
+  upload-targets:
+    type: array
+    items:
+      type: object
+      oneOf:
+        - properties:
+            name:
+              type: string
+            method:
+              enum: ["google-cloud-storage"]
+          required: ["name", "method"]
+          additionalProperties: false
+        - properties:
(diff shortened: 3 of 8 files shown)


krawthekrow:
For reference, here are my setup notes:

```
[install development vm]

cd ~
mkdir homeworld-vm
cd homeworld-vm
mkdir images
(cd images && wget https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-9.4.0-amd64-netinst.iso)
mkdir storage
qemu-img create -f qcow2 storage/homeworld-deploy.qcow2 20G
virt-install --name homeworld-deploy --ram 1024 --disk path=storage/homeworld-deploy.qcow2,size=20 --vcpus 1 --os-type linux --os-variant generic --network network=default --graphics spice -c images/debian-9.4.0-amd64-netinst.iso
[install]

[in vm]
su
apt install sudo ssh-server nginx rsync
adduser <user> sudo
reboot

[in local]
virsh net-dhcp-leases default
ssh-copy-id <user>@<ip>
sftp <user>@<ip>
[put homeworld-apt-setup.deb]

[in vm]
sudo echo '127.0.0.1 homeworld.deploy' > /etc/hosts
cd /var/www/html

[in local chroot]
ssh-keygen -t rsa
ssh-copy-id <user>@<ip>

[in vm]
[disable ssh password authentication]
sudo chown <user> /var/www/html
sudo chgrp <user> /var/www/html
mkdir -p /var/www/html/<apt branch>

[in local chroot]
glass --upload

[in vm]
sudo dpkg -i <apt setup deb>
sudo apt update
sudo apt install homeworld-admin-tools
```

building/glass/aptbranch.py
```diff
@@ -4,7 +4,11 @@
 import re
 import subprocess
 
+import validate
+
 SETUP_DIR = "/h/setup-apt-branch/"
+CONFIG_PATH = "/h/apt-branch-config/branches.yaml"
+CONFIG_SCHEMA_NAME = "branches-schema.yaml"
```
celskeggs: This is a great change to branches (at very first glance); please submit it first and separately.

building/glass/upload.py
```diff
@@ -122,20 +121,27 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
+            if not os.path.exists(BOTO_PATH):
+                raise Exception("you need to put the GCP service account private key file into /homeworld/boto-key")
+            botoconfig_name = os.path.join(staging, "boto.config")
+            with open(botoconfig_name, "w") as bout:
+                with open(BOTO_PATH, "r") as f:
+                    project_id = json.load(f)["project_id"]
+                bout.write(BOTO_TEMPLATE % (BOTO_PATH, project_id))
+                bout.flush()
+            gs_rsync(root, branch_config.name, botoconfig_name)
+        elif upload_method == "rsync":
```
celskeggs: This should dispatch to different methods, rather than using an if/else sequence.

building/glass/upload.py
```diff
@@ -122,20 +121,27 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
+            subprocess.check_call(["rsync", "-a", root + "/", "%s@%s:%s" % (branch_config.upload_config["username"], branch_config.upload_config["host"], os.path.join(branch_config.upload_config["dir"], branch_config.name))])
+        else:
+            raise Exception("unrecognized upload method")
```
celskeggs: You should include the name of the unrecognized method.

celskeggs:
Looks pretty good at first glance (see comments); I'm a bit unsure about the idea of deploying from the same VM that we upload downloadable binaries to?

(I'm not actually sure we want a dedicated VM to serve binaries at all?)

krawthekrow:
My plan was indeed to serve the binaries from the same VM, just to conserve resources since it's a local deploy. But the requested changes should still apply even if the binaries are served from elsewhere. Nevertheless, I do agree that serving the binaries from a separate machine is the more correct approach.

celskeggs:
Two other options (up to your discretion) are:

- We could add a feature to `spire` that launches a simple webserver to serve certain binaries.
- We could serve the binaries directly off of the host.

krawthekrow:
I have tested this on a VPS, so this should really be agnostic to where the user chooses to serve the binaries. I guess this PR isn't really a step towards local deploy in general, it just allows us to serve binaries without having access to GCS. We could continue discussing the details of the local deploy process under the corresponding issue (which doesn't seem to be created yet? Correct me if I'm wrong).

Specific to this PR, we could even allow the binaries to be served from AFS, for example by allowing the username and host to be omitted.

building/apt-branch-config/branches.yaml.example
```diff
@@ -4,3 +4,8 @@ branches:
   - name: <name>/<subbranch>
     signing-key: <signing key>
     apt-url-prefix: homeworld.deploy
+    upload:
+      method: rsync
+      username: <username>
+      host: <ip address>
```
celskeggs: Not sure it makes sense to imply that this is limited to IP addresses?

building/glass/branches-schema.yaml
```diff
@@ -15,6 +15,24 @@ properties:
...
+          - properties:
+              method:
+                enum: ["rsync"]
+              username:
+                type: string
+              host:
+                type: string
+              dir:
+                type: string
+            required: ["username", "host", "dir"]
```
celskeggs: I think you need to include `method` as a required field?

building/glass/upload.py
```diff
@@ -123,20 +121,40 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
+        with open(BOTO_PATH, "r") as f:
+            project_id = json.load(f)["project_id"]
+        bout.write(BOTO_TEMPLATE % (BOTO_PATH, project_id))
+        bout.flush()
+    gs_rsync(root, branch, botoconfig_name)
+
+
+def upload_rsync(staging, root: str, branch_config: aptbranch.Config):
+    target_dir = os.path.join(branch_config.upload_config["dir"], branch_config.name)
+    target = "%s@%s:%s" % (branch_config.upload_config["username"], branch_config.upload_config["host"], target_dir)
```
celskeggs: It might be worth erroring out here if any special characters are included in these fields that could be problematic for rsync's parsing of its parameter? For example, a `-` at the start of the username.

building/glass/upload.py
```diff
@@ -122,20 +121,40 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
-                project_id = json.load(f)["project_id"]
-            bout.write(BOTO_TEMPLATE % (BOTO_PATH, project_id))
-            bout.flush()
-        gs_rsync(root, branch, botoconfig_name)
+
+        upload_method = branch_config.upload_config["method"]
+        if upload_method == "google-cloud-storage":
+            upload_gs(staging, root, branch_config)
+        elif upload_method == "rsync":
+            upload_rsync(staging, root, branch_config)
```
celskeggs: Maybe use a dictionary of upload methods, which maps "possible upload method" to "uploading function for that upload method"?

celskeggs:
Local deploy issue (now created): #295.

building/glass/upload.py
krawthekrow: I've added "--", which should take care of the dash problem. I also combined the username and host in the config and checked if it contains a colon. Is there anything I missed?

krawthekrow:
I added a commit to copy the user's SSH identity into the chroot, since it would be needed for rsync.

building/glass/upload.py
celskeggs: I don't know. I usually try to use a whitelist instead of a blacklist, because trying to blacklist format violations is generally a losing game.

However, since we can trust that anyone who can modify the branches.yaml is probably not evil anyway, this might not matter.

building/glass/upload.py
```diff
@@ -123,20 +119,43 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
+        bout.flush()
+    gs_rsync(root, branch, botoconfig_name)
+
+
+def upload_rsync(staging, root: str, branch_config: aptbranch.Config):
+    target_dir = os.path.join(branch_config.upload_config["dir"], branch_config.name)
+    if ":" in branch_config.upload_config["target"]:
+        raise Exception("colon in rsync upload target")
+    dest = "%s:%s" % (branch_config.upload_config["target"], target_dir)
+    subprocess.check_call(["rsync", "-av", "--progress", "--delete-delay", "--", root + "/", dest])
```
celskeggs: Maybe you want `-z` and/or `-c` as well?

building/glass/upload.py
```diff
@@ -123,20 +119,43 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
-            with open(BOTO_PATH, "r") as f:
-                project_id = json.load(f)["project_id"]
-            bout.write(BOTO_TEMPLATE % (BOTO_PATH, project_id))
-            bout.flush()
-        gs_rsync(root, branch, botoconfig_name)
+
+        upload_funcs = {
+            "google-cloud-storage": upload_gs,
+            "rsync": upload_rsync
+        }
```
celskeggs: I would define this at the top level, near where the functions it references are defined.

enter-chroot.sh
```diff
@@ -23,4 +23,9 @@ then
 	cp "$HOME/.gnupg/trustdb.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/trustdb.gpg"
 	cp -R "$HOME/.gnupg/private-keys-v1.d/"* "$HOMEWORLD_CHROOT/home/$USER/.gnupg/private-keys-v1.d"
 fi
+if [ -e "$HOME/.ssh" ]
+then
+	mkdir -p "$HOMEWORLD_CHROOT/home/$USER/.ssh/"
+	cp "$HOME/.ssh/id_rsa" "$HOMEWORLD_CHROOT/home/$USER/.ssh/id_rsa"
```
celskeggs: What if they use, for example, an ecsda key, or something else?

Also, your commit message is misspelled.

building/glass/aptbranch.py
```diff
@@ -51,11 +51,20 @@ def select_branch_config(branches_config: list, branch: str):
...
+
+
 class Config:
     def __init__(self, branch: str):
         if not os.path.exists(CONFIG_PATH):
             raise Exception("cannot find branches config at %s, use %s.example to create one" % (CONFIG_PATH, CONFIG_PATH))
-        branches_config = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)["branches"]
+        configs = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)
+        branches_config = configs["branches"]
+        self.upload_targets = configs.get("upload-targets")
```
celskeggs: Why not just `configs.get("upload-targets", [])` instead of using get like this and then checking for None later?

celskeggs:
Please rebase and fix tests so that I can review again.

celskeggs:
Once this has been fully rebased, please assign it to me so that I get a notification.

celskeggs:
Looks good. Have you tested a full deploy? Are you ready to merge?

celskeggs:
You will also need to rebase this again, sorry.

celskeggs:
ping @krawthekrow

building/glass/upload.py
```diff
@@ -123,20 +119,47 @@ def gs_rsync(local_path: str, remote_path: str, boto_path: str):
...
+        with open(BOTO_PATH, "r") as f:
+            project_id = json.load(f)["project_id"]
+        bout.write(BOTO_TEMPLATE % (BOTO_PATH, project_id))
+        bout.flush()
+    gs_rsync(root, branch, botoconfig_name)
+
+
+def upload_rsync(staging, root: str, branch_config: aptbranch.Config):
+    branch_name_parts = branch_config.name.split('/', 1)
+    target_dir = os.path.join(branch_config.upload_config["dir"], branch_name_parts[1])
```
celskeggs: Maybe use unpacking, instead of branch_name_parts?

i.e. `host, path = branch_config.name.split('/', 1)`

enter-chroot.sh
```diff
@@ -24,4 +24,9 @@ then
 	cp "$HOME/.gnupg/trustdb.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/trustdb.gpg"
 	cp -R "$HOME/.gnupg/private-keys-v1.d/"* "$HOMEWORLD_CHROOT/home/$USER/.gnupg/private-keys-v1.d"
 fi
+if [ -e "$HOME/.ssh" ]
+then
+	mkdir -p "$HOMEWORLD_CHROOT/home/$USER/.ssh/"
+	cp -R "$HOME/.ssh/"* "$HOMEWORLD_CHROOT/home/$USER/.ssh"
+fi
```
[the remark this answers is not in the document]
krawthekrow: Would `cp -R "$HOME/.ssh/." "$HOMEWORLD_CHROOT/home/$USER/.ssh"` be better? Line 25 of the same file has the same problem.

krawthekrow:
I don't think the changes I made (the only changes I made were to directly address the review) should affect whether it deploys successfully, so I would like it to pass review before I run it through another deploy.

PR #293, opened 2018-07-06
Set up configuration for apt branches

krawthekrow:
Just the configuration file portion from #292 (with some cleanup).


building/glass/aptbranch.py
```diff
@@ -23,46 +26,13 @@ def get_env_branch():
...
     "Get the signing key ID for a specific apt branch. Throw an exception if not found, or if the key doesn't exist in the gpg keyring."
 
-    key_id = get_signing_key_unchecked(branch)
     if subprocess.call(["gpg", "--list-keys", key_id], stdout=subprocess.DEVNULL) != 0:
         if branch == "root/master":
             print("If you're basing this build off the master branch, import its signing key with")
-            print('gpg --import building/setup-apt-branch/default-key.asc')
+            print('gpg --import building/apt-branch-config/default-key.asc')
         raise Exception("apt signing key not in gpg keyring: %s" % branch)
     return key_id
```
celskeggs: No need to return the key id anymore.

building/glass/aptbranch.py
```diff
@@ -4,7 +4,10 @@
 import re
 import subprocess
```
celskeggs: Extra newline?

building/glass/aptbranch.py
```diff
@@ -23,46 +26,13 @@ def get_env_branch():
...
-            signing_key = kvs[branch]
-            if not re.match("^[0-9a-zA-Z]+$", signing_key):
-                raise Exception("apt signing key invalid: %s" % signing_key)
-            return signing_key
-    raise Exception("apt branch %s not found in signing keys" % branch)
-
-
-def get_signing_key(branch):
+def check_signing_key(key_id):
     "Get the signing key ID for a specific apt branch. Throw an exception if not found, or if the key doesn't exist in the gpg keyring."
```
celskeggs: Update -- or at least remove -- this docstring.

building/glass/aptbranch.py
```diff
@@ -73,3 +43,34 @@ def export_key(keyid, armor=False):
     if not result.strip():
         raise Exception("empty result from gpg for keyid: '%s'" % keyid)
     return result
+
+
+class AptBranchConfig:
+    def __init__(self, branch: str):
+        if not os.path.exists(CONFIG_PATH):
+            raise Exception("cannot find apt branch config under %s" % CONFIG_PATH)
```
celskeggs: Maybe move this exception into load_validated?

building/glass/aptbranch.py
```diff
@@ -73,3 +43,34 @@ def export_key(keyid, armor=False):
...
+        branches_config = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)
+        self.config = None
+        for branch_config in branches_config["branches"]:
+            if branch_config["name"] == branch:
+                self.config = branch_config
+                break
+
+        if self.config is None:
+            raise Exception("no config found for %s in %s" % (branch, CONFIG_PATH))
```
celskeggs: I generally prefer assigning each instance field once during __init__, by convention. I'd have a separate 'config' variable, and only assign it to this.config once it's been fully determined.

Maybe even separate this part out into a separate function.

This is up to your discretion.

building/glass/aptbranch.py
```diff
@@ -73,3 +43,34 @@ def export_key(keyid, armor=False):
...
+
+    @property
+    def signing_key(self) -> str:
+        return self.config["signing-key"]
+
+    @property
+    def apt_url(self):
+        if "apt-url-prefix" in self.config:
+            return self.config["apt-url-prefix"] + self.name
+        return self.name
```
celskeggs: To make this functionality more obvious, I would either include an `else` clause or use the pattern of 'assign to local variable; mutate variable if condition holds; return local variable'

building/glass/branches-schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+type: object
+properties:
+  branches:
+    type: array
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+          pattern: "[a-z0-9.-]+/[a-z0-9-]+"
```
celskeggs: Maybe we should also support underscores in this?

building/glass/branches-schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+          - properties:
+              method:
+                enum: ["rsync"]
+              username:
+                type: string
+              host:
+                type: string
+              dir:
+                type: string
+      required: ["name", "signing-key"]
```
celskeggs: Should probably disallow other unrecognized fields on all `type: object` schemas, so that typos are more likely to be caught.

building/glass/glass.py
```diff
@@ -43,16 +45,16 @@
     if args.clean or args.rebuild:
         any = False
         for p in projects:
-            any |= p.clean(args.branch)
+            any |= p.clean(branch_config.name)
```
celskeggs: It feels a little bit weird to pass a branch config name to `clean`, and a branch config to `run`. Maybe these should both take a full config?

building/glass/aptbranch.py
```diff
@@ -73,3 +43,34 @@ def export_key(keyid, armor=False):
     if not result.strip():
         raise Exception("empty result from gpg for keyid: '%s'" % keyid)
     return result
+
+
+class AptBranchConfig:
```
celskeggs: Maybe just call this `Config`? It's already implied to be an _apt branch_ config by way of the module that it's in.

building/glass/aptbranch.py
krawthekrow: This separates imports for our own files from stdlib imports, as per convention.

celskeggs: Woops, yeah.

building/glass/aptbranch.py
```diff
@@ -73,3 +42,41 @@ def export_key(keyid, armor=False):
...
+
+class Config:
+    def __init__(self, branch: str):
+        if not os.path.exists(CONFIG_PATH):
+            raise Exception("cannot find branches config at %s, use %s.example to create one" % (CONFIG_PATH, CONFIG_PATH))
+        branches_config = validate.load_validated(CONFIG_PATH, CONFIG_SCHEMA_NAME)["branches"]
+        self.config = select_branch_config(branches_config, branch)
+
+        if self.config is None:
+            raise Exception("no config found for %s in %s" % (branch, CONFIG_PATH))
```
celskeggs: Maybe it makes more sense to just directly throw this, rather than returning None and then checking for None?

building/glass/aptbranch.py
```diff
@@ -73,3 +42,41 @@ def export_key(keyid, armor=False):
...
+        return self.config["signing-key"]
+
+    @property
+    def apt_url(self):
+        if "apt-url-prefix" in self.config:
+            url_prefix = self.config["apt-url-prefix"]
+            if not url_prefix.endswith("/"):
+                url_prefix += "/"
+        else:
+            url_prefix = ""
```
celskeggs: May I suggest: `self.config.get("apt-url-prefix", "")`?

building/glass/branches-schema.yaml
```diff
@@ -0,0 +1,21 @@
...
+type: object
+properties:
+  branches:
+    type: array
+    items:
+      type: object
+      properties:
+        name:
+          type: string
+          pattern: "[a-z0-9._-]+/[a-z0-9_-]+"
```
celskeggs: Why aren't we supporting branches with multiple slashes, or dots after the last slash? (I don't remember exactly, but I think we implicitly supported those previously.)

celskeggs: Ignore this nitpick.

krawthekrow: Good catch with the period, though. I had royally messed up the regexes in this file in more ways than one.

krawthekrow:
Please do continue nitpicking, your points are very valid.

building/glass/aptbranch.py
```diff
@@ -73,3 +42,35 @@ def export_key(keyid, armor=False):
...
+
+    @property
+    def signing_key(self) -> str:
+        return self.config["signing-key"]
+
+    @property
+    def apt_url(self):
+        url_prefix = self.config.get("apt-url-prefix", "")
+        if url_prefix != "" and not url_prefix.endswith("/"):
+            url_prefix += "/"
```
celskeggs: You can just do `url_prefix and ...` instead of `url_prefix != "" and ...` (non-empty strings evaluate to true)

celskeggs:
Oh, also, it might be worth making a better commit message here.

Maybe something like:

`upload: convert branch config to yaml`

followed by an appropriate explanation of the point of the change.

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           command: echo "make -C upstream-check/ -j2 verify && killall gpg-agent" | HOMEWORLD_CHROOT="$HOME/autobuild-chroot" USER="circleci" ./enter-chroot-ci.sh
       - run:
           name: Launch build with glass
-          command: echo "glass components -b root/master" | HOMEWORLD_CHROOT="$HOME/autobuild-chroot" USER="circleci" ./enter-chroot-ci.sh
+          command: echo "glass components -b root/master --branches-config /h/apt-branch-config/branches-ci.yaml" | HOMEWORLD_CHROOT="$HOME/autobuild-chroot" USER="circleci" ./enter-chroot-ci.sh
```
celskeggs: Why not just copy branches-ci.yaml file over the standard location? The point is to test something close to what users will actually do.

krawthekrow: That's true, there isn't really a need to use a different branches config path.

celskeggs:
Actually, re: your commit message -- the part before the colon is supposed to be a single word or filename, and not have any spaces. also 'apt branch config' and 'apt branches' are redundant. I recommend 'upload:'

krawthekrow:
I'm reluctant to put "upload" there since it isn't technically part of the upload process, so I'll go with "glass".

celskeggs:
Very good point.

PR #296, opened 2018-07-10
chroot: add snapshot.debian.org cache

### LINKED ISSUE #286: Cache debian packages in upstream repository
Related to #102, it would be very helpful for build speed if we didn't need to redownload debian packages from the archive every time.

celskeggs:
Fixes #286.

We rely on using snapshot.debian.org for reproducibility.
Unfortunately, it is also very slow, and this has a severe effect on
the speed of our build process. In order to improve build performance,
we need to cache downloaded debian packages in our upstream directory.

Add the nginx configuration required for this cache and add the scripts required to download the cache into the upstream directory.


celskeggs:
From empirical testing, this lowers the time to build the `debian` component from 644 seconds to 213 seconds.

building/upstream-check/snapshot-pull.py
```diff
@@ -0,0 +1,55 @@
...
+
+if len(sys.argv) != 2 or sys.argv[1][0] != "2" or sys.argv[1][-1] != "Z":
+	print("usage: snapshot-pull.py <snapshot-version>")
+	print("   version example: 20180710T043017Z")
+	sys.exit(1)
+
+VERSION = sys.argv[1]
+prefix = "snapshot.debian.org/archive/debian/%s" % VERSION
+
+def download(path):
```
krawthekrow: Why not use an existing solution to mirror the repository? Also, since apt_pkg is a dependency anyway, why not use it to parse the package files?

enter-chroot.sh
```diff
@@ -23,4 +23,6 @@ then
 	cp "$HOME/.gnupg/trustdb.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/trustdb.gpg"
 	cp -R "$HOME/.gnupg/private-keys-v1.d/"* "$HOMEWORLD_CHROOT/home/$USER/.gnupg/private-keys-v1.d"
 fi
-sudo systemd-nspawn --bind $(pwd):/homeworld:norbind -u "$USER" -a -D "$HOMEWORLD_CHROOT" bash -c "cd /h/${ORIG_REL} && gpg-agent --daemon --keep-tty && exec bash"
+sudo cp "$HOMEWORLD_CHROOT/etc/hosts.template" "$HOMEWORLD_CHROOT/etc/hosts"
+(echo "127.0.0.1 $(basename "$HOMEWORLD_CHROOT")"; echo "127.0.0.2 snapshot.debian.org") | sudo bash -c "cat >>'$HOMEWORLD_CHROOT/etc/hosts'"
```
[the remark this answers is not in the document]
celskeggs: That's actually a very good question.

enter-chroot-ci.sh
```diff
@@ -20,10 +20,14 @@ then
 	mkdir -p "$HOMEWORLD_CHROOT/home/$USER/.gnupg"
 	cp "$HOME/.gnupg/pubring.gpg" "$HOMEWORLD_CHROOT/home/$USER/.gnupg/pubring.gpg"
 fi
+sudo cp "$HOMEWORLD_CHROOT/etc/hosts.template" "$HOMEWORLD_CHROOT/etc/hosts"
+(echo "127.0.0.1 $(basename "$HOMEWORLD_CHROOT")"; echo "127.0.0.2 snapshot.debian.org") | sudo bash -c "cat >>'$HOMEWORLD_CHROOT/etc/hosts'"
```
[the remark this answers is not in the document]
krawthekrow: I don't think this works if, say, there's a space in the chroot directory name, right? Maybe you should check for that and error out in enter-chroot.sh.

building/upstream-check/snapshot-pull.py
celskeggs: I can't find good documentation for apt_pkg, and it doesn't look like it would be straightforward to integrate. I'd rather stick with this code for now and leave migration for later.

create-chroot.sh
```diff
@@ -11,6 +11,12 @@ then
     exit 1
 fi
 
+if [ "${HOMEWORLD_CHROOT}" =~ " " ]
```
krawthekrow: Do you mean $(basename "${HOMEWORLD_CHROOT}")? It's fine if there are spaces in ancestor directories.

celskeggs: It's probably a bad thing for there to be spaces in ancestor directories.

Might it work? Yeah.

But spaces tend to wreck everything, so this is safer.

celskeggs:
(I also fixed an unrelated typo in create-chroot.sh that was causing tests to fail.)

create-chroot.sh
krawthekrow: I don't think we should be setting constraints on the user's wider system, but this matter is minor enough that I won't let it block the PR.

celskeggs: "don't use spaces" is a fairly normal constraint on a user's system.

PR #297, opened 2018-07-14
debian: bump installer version for 9.5.0

celskeggs:
Debian updated its install CD, which means they stopped serving the old
install CD. We need to bump the version of the installer that we use
accordingly, so that our CI can pull down the correct ISO version.


celskeggs:
@krawthekrow this is more important to review than #296

celskeggs:
@krawthekrow: I think this should be good; okay for me to merge?

celskeggs:
Ignore the flurry of mistakes I just made. I repushed the correct commits.

PR #298, opened 2018-07-15
 cluster: bump container versions

celskeggs:
cluster: bump container versions
    
    Bump the versions of dnsmasq-nanny and flannel in the kubernetes
    configuration files.
    
    The previous commit bf84c1f4ba68f97c93e96dc002d0d7b79fcbb929 updated
    the versions of the dnsmasq-nanny and flannel containers as a result of
    updating their debian versions, but failed to also update the versions
    used for cluster deploy itself.

This is one of those times when I regret not testing previous changes more thoroughly.

~(This also contains a commit from the previously-approved PR #296, because I'm still waiting on that PR's tests passing before merge; please ignore that.)~

### DIFF
--- a/building/components/homeworld-admin-tools/resources/clustered/dns-addon.yaml
+++ b/building/components/homeworld-admin-tools/resources/clustered/dns-addon.yaml
@@ -87,7 +87,7 @@ spec:
           name: metrics
           protocol: TCP
       - name: dnsmasq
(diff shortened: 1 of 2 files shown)


celskeggs:
@krawthekrow: rereview?

celskeggs:
Performed a local test and deployment worked correctly.

PR #302, opened 2018-07-15
spire: eliminate cdpack system, split out ISO into separate package

celskeggs:
Remove the separate step of extracting the debian ISO during build, and
eliminate the effort of rebuilding and reuploading a copy of the debian
ISO every time that spire is modified.

The spire cdpack system was implemented when we didn't have bsdtar, and
correspondingly needed to perform an actual mount operation to read the
files out of the debian base ISO. We no longer need to do this, so the
cdpack process can be re-integrated back into the main ISO remastering
process.

We also relocate the debian ISO to /usr/share/homeworld/debian.iso, because
it doesn't make much sense to incorporate this kind of binary into the
deployed spire binary, when we could also put it in the debian package -- or
even a different package.

### DIFF
--- a/building/components/homeworld-admin-tools/glass.yaml
+++ b/building/components/homeworld-admin-tools/glass.yaml
@@ -1,6 +1,6 @@
 control:
   name: homeworld-admin-tools
-  version: 0.2.3
+  version: 0.2.4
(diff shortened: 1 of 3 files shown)


celskeggs:
Merging once CI passes.

PR #304, opened 2018-07-18
spire, upstream, chroot: miscellaneous improvements

celskeggs:
See commit descriptions. These changes passed testing in a dev cluster.


ichung:
These changes are fine, but I notice there's some suspect things about the surrounding code, specifically the `IP` class:
- implementation of __eq__, __ne__ are incorrect because objects of different types should compare inequal
- implementation of __and__ should raise an error instead of returning one
- The whole class could probably be replaced with uses of the [ipaddress](https://docs.python.org/3/library/ipaddress.html) standard library module

celskeggs:
I've split out #307 as an issue for that. Feel free to assign yourself if you feel like it.

PR #312, opened 2018-08-01
install: prepare for automatic installation

celskeggs:
Modify our install process to be more easily automated.

 - Allow configuring mirrors
 - Move registry from homeworld.mit.edu to homeworld
 - Add custom root CA and grant our own cert from it
 - Bypass DNS for hostname and kerberos realm config
 - Allow installing without kerberos
 - Set up for prefillable IP address and bootstrap token from isolinux boot prompt

---

Checklist:

 - [ ] My changes have passed CI.


building/components/homeworld-admin-tools/src/configuration.py
```diff
@@ -173,6 +173,9 @@ def __init__(self, kv: dict):
                     command.fail("in config: multiple supervisors not yet supported")
                 self.keyserver = node
 
+    def is_kerberos_enabled(self):
```
ichung: I'm a little confused about this computation.  What does the `root_admins` key usually mean?  Also, how exactly does the disabled-kerberos configuration work?

celskeggs: The `root_admins` key refers to the list of kerberos principals that can authenticate as cluster root. If none are registered, we don't need kerberos auth.

Disabled kerberos configuration means no keygateway, no checking that the keygateway works, no keytab uploading, and needing to grant ssh certificates in a different way.

ichung: How are ssh certs granted in this case?

celskeggs:
See https://github.com/sipb/homeworld/commit/91e081ecc9a3bdd742f7edabef5fc594e6f7daef

On Aug 5, 2018 20:27, ikdc <notifications@github.com> wrote:

@ichung commented on this pull request.

________________________________

In building/components/homeworld-admin-tools/src/configuration.py<https://github.com/sipb/homeworld/pull/312#discussion_r207766938>:

> @@ -173,6 +173,9 @@ def __init__(self, kv: dict):
                     command.fail("in config: multiple supervisors not yet supported")
                 self.keyserver = node

+    def is_kerberos_enabled(self):


How are ssh certs granted in this case?

ichung:
Is this tested in configurations with kerberos {enabled, disabled}?

building/components/homeworld-admin-tools/src/configuration.py
celskeggs: See https://github.com/sipb/homeworld/commit/91e081ecc9a3bdd742f7edabef5fc594e6f7daef

[moved here from other incorrectly-added email comment]

celskeggs:
This has not been tested on the official dev cluster. I'm waiting for review to finish, so that I don't have to re-test.

celskeggs:
@ichung: I'm testing a newer version, which means I need to add six more commits. Re-review to include these? (You shouldn't have to re-review the changes you've already reviewed.)

celskeggs:
Rebased to do homeworld.mit.edu -> homeworld.private instead of homeworld.mit.edu -> homeworld, because of issues discovered during testing about how kubernetes handles TLD-less domain names.

building/components/homeworld-admin-tools/src/access.py
```diff
@@ -241,11 +242,50 @@ def ssh_foreach(ops: setup.Operations, node_kind: str, *params: str):
...
+
+
+def hostkeys_by_fingerprint(node: configuration.Node, fingerprints: list):
+    keys = []
+    for line in subprocess.check_output(["ssh-keyscan", "-T", "1", "--", str(node.ip)]).decode().split("\n"):
+        if not line or line.startswith("#"): continue
+        assert line.count(" ") >= 1
+        server, key = line.split(" ", 1)
+        assert server == str(node.ip)
+        fingerprint = compute_fingerprint(key)
```
ichung: Is there a reason to remove the comment (here, server name) so that it doesn't appear in the fingerprint, but the resulting non-comment comment (manifesting as the text `no comment`) is left in the fingerprint itself?  I think if we don't consider the comment part of the fingerprint then we should just strip it out of the fingerprint after we pass it through `ssh-keygen -lf`.  Or if it is part of the fingerprint then we should leave it intact and pass the real one to `ssh-keygen -lf`.

building/components/homeworld-admin-tools/src/access.py
```diff
@@ -241,11 +242,50 @@ def ssh_foreach(ops: setup.Operations, node_kind: str, *params: str):
...
+        subprocess.check_call(["ssh-keygen", "-f", known_hosts, "-R", remove],
+                              stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+
+    keys = hostkeys_by_fingerprint(node, fingerprints)
+    with open(known_hosts, "a") as f:
+        for key in keys:
+            f.write("%s.%s %s\n" % (node.hostname, config.external_domain, key))
+
+
+def pull_supervisor_key_from(source_file):
```
ichung: Where do we get this file from?

building/components/homeworld-admin-tools/src/access.py
```diff
@@ -241,11 +242,50 @@ def ssh_foreach(ops: setup.Operations, node_kind: str, *params: str):
...
 
 
+def compute_fingerprint(key: str) -> str:
+    text = subprocess.check_output(["ssh-keygen", "-l", "-f", "-"], input=key.encode(), stderr=subprocess.STDOUT).decode()
+    assert text.endswith("\n") and text.count("\n") == 1
+    return text.rstrip("\n")
+
+
+def hostkeys_by_fingerprint(node: configuration.Node, fingerprints: list):
+    keys = []
```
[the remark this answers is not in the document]
celskeggs: No. N = 3.

celskeggs:
Great! Just waiting on CI now.

Issue #320, opened 2018-08-20
Wrong trusted key provided to rkt


krawthekrow:
Currently, the trusted key used for rkt signature verification is copied directly from a file in homeworld-services ([here](https://github.com/sipb/homeworld/blob/fb85815c050cb0b18bd4bfaabb0d30ff0f5491df/building/components/homeworld-services/glass.yaml#L37-L39)). This ought to be the key used to sign the repositories, that is, the one indicated in branches.yaml.

celskeggs:
Weird. I thought I'd fixed that.

celskeggs:
Fixed by #321.

PR #321, opened 2018-08-21
services: Provide the correct repository signing key to rkt

krawthekrow:
Previously, a fixed public key was provided to rkt for ACI signature
verification, preventing developers without the corresponding private
key from deploying the cluster. Here, we instead extract the key from
GPG, as is done in homeworld-apt-setup.

The implementation is rather kludgy since glass doesn't provide a native
way to write to a file with a dynamically determined name (rkt expects
the name of the file to be the fingerprint of the key). Any attempts at
generalization (e.g. modifying the build step) is deferred to when
similar features are required elsewhere.

This is built over commits from #292 since I needed them to deploy.

See #320.

---

Checklist:

 - [ ] My changes have passed CI.


celskeggs:
Please rebase this against your other changes in #292, and perform a cluster deploy. I'll merge both PRs at once when they've been tested.

PR #325, opened 2018-09-01
Unify and clean up compilation of golang source code

celskeggs:
Refactor all of our golang source code into a single unified tree. This allows us to eliminate the sources-shared hack and simplify some of our glassfiles, and prepares us for further, more comprehensive simplifications to our golang builds in the future.

Don't try to review this all at once; the individual commits should be well-delineated, and the messages should explain why each change is being made.

---

Checklist:


celskeggs:
@krawthekrow: when you get the chance, can you take a look at this?

In addition to normal review commentary, it'd be good to know your opinion on whether this actually makes things simpler. I think it does, but I want a second opinion.

Once this passes review, I'll do a full deploy cycle.

krawthekrow:
I think it makes things simpler from a building perspective. From a conceptual perspective:

- It's not clear where "pull" belongs to. If it's only used by flannel-monitor, it should be made more explicit, perhaps by putting it inside the flannel-monitor folder in the source directory.
- It's less clear now that keycommon is shared outside keysystem. I imagine that if I were a new developer and edited code in keycommon, I would be surprised when things other than the keysystem break. Maybe they could be refactored out to a separate util module?
- Naming the go src directory "src" makes it seem like it contains all the source code for the entire project. "gosrc" or similar might be better.
- Not immediately important, but the keysystem integration tests still live in the homeworld-keysystem directory. Have you checked that they still pass? Eventually, these should be moved to the unified go directory too.

Other than that, it looks like an improvement to me.

celskeggs:
- I've made the flannel-monitor situation clearer.
- This code is too usage-specific for it to make sense to me to make it a utility package. Maybe I could rename it from keysystem/keycommon to keysystem/api? That might make it more obvious that it was used more broadly.
- I'm intentionally calling it 'src' instead of 'gosrc' because I want to eventually move other code into the same directory, as well. git never seems to enjoy bulk renames, so I wanted to avoid calling something 'gosrc' if I could just call it 'src'. What do you think?
 - I've split this out into #326.

krawthekrow:
- Yes, "api" would be a better name for it.
- That's reasonable, but then we would have separate utils for go and python, though we can defer thinking about this until it actually becomes a problem.

PR #349, opened 2018-12-14
 Change format of apt branches. 

ichung:
Separate the following concepts from each other:
- name of branch
- download URL
- upload method

This allows for a simpler and more flexible model of branches.  I haven't tested this with GCS upload, only rsync.

---

Checklist:

 - [ ] [My changes have passed CI.]  CI is down


chroot-packages.list
```diff
@@ -32,7 +32,7 @@ bzip2, ca-certificates, curl, gzip, less, sudo
 python3-jsonschema, reprepro, python3-apt
 
 # uploading
-python-crcmod, python-pip, python-setuptools, python-wheel, ssh
+python-crcmod, python-pip, python-setuptools, python-wheel, rsync, ssh
```
krawthekrow: This file doesn't seem to repeat dependencies that were already listed earlier. If this is a convention we're following, then this is redundant.

building/glass/branches-schema.yaml
```diff
@@ -2,42 +2,37 @@ $schema: "http://json-schema.org/draft-07/schema#"
...
     items:
       type: object
       properties:
         name:
           type: string
-          pattern: "^[0-9a-zA-Z_.-][0-9a-zA-Z_./-]+[0-9a-zA-Z_.-]$"
+          pattern: "[0-9a-zA-Z_-]+"
+        download:
+          type: string
+          pattern: "^no_download$|^https?://[0-9a-zA-Z_.-]+(:[0-9]+)?(/[0-9a-zA-Z_./-]*[0-9a-zA-Z_.-])?$"
```
krawthekrow: It might be better to just leave the download field blank if no download is intended, and fill in the fake value in Python. Otherwise, this field should be marked required.

building/glass/branches-schema.yaml
```diff
@@ -2,42 +2,37 @@ $schema: "http://json-schema.org/draft-07/schema#"
...
-        upload-target:
-          type: string
+        upload:
+          type: object
+          oneOf:
+            - properties:
+                method:
+                  enum: ["google-cloud-storage"]
+                gcs-target:
+                  type: string
```
krawthekrow: Might be good to validate the gcs-target and rsync-target fields.

chroot-packages.list
ichung: imo we should list all dependencies required for each task, since the way we accomplish each task might change

building/glass/branches-schema.yaml
ichung: that is a bit difficult since rsync targets can have many different forms.  I think it's not worth it, especially since if it's wrong then the rsync command will fail quickly enough anyway

building/glass/branches-schema.yaml
ichung: marked required since "no download" is an exceptional case for CI only

ichung:
@celskeggs I haven't tested this branch on GCS since i don't have an account with them right now.  Do you want to check if uploading still works with this branch before it's merged or just merge and find out later?

building/glass/branches-schema.yaml
```diff
@@ -2,43 +2,38 @@ $schema: "http://json-schema.org/draft-07/schema#"
...
-          additionalProperties: false
   branches:
     type: array
     items:
       type: object
       properties:
         name:
           type: string
-          pattern: "^[0-9a-zA-Z_.-][0-9a-zA-Z_./-]+[0-9a-zA-Z_.-]$"
+          pattern: "[0-9a-zA-Z_-]+"
```
celskeggs: Is there a reason that the anchors were removed from this expression?

building/glass/branches-schema.yaml
```diff
@@ -2,43 +2,38 @@ $schema: "http://json-schema.org/draft-07/schema#"
...
-      required: ["name", "signing-key"]
+        upload:
+          type: object
+          oneOf:
+            - properties:
+                method:
+                  enum: ["google-cloud-storage"]
+                gcs-target:
+                  type: string
+              required: ["gcs-target"]
```
celskeggs: I think you might need to include "method" in the required array?

building/glass/branches-schema.yaml
ichung: oops

ichung:
@celskeggs re-review?

building/glass/aptbranch.py
```diff
@@ -89,9 +71,10 @@ def name(self) -> str:
...
+    def download(self) -> str:
+        return self.config["download"]
+
     @property
     def upload_config(self):
-        if "upload-target" not in self.config:
-            raise Exception("no upload target found for branch %s" % self.name)
-        upload_target_name = self.config["upload-target"]
-        return select_upload_config(self.upload_targets, upload_target_name)
+        return self.config["upload"]
```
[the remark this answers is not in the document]
ichung: This is the case for our current CircleCI setup, so it's not required.  In that case it's an error to try and upload with glass.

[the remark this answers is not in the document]
celskeggs: Can we preserve the behavior of giving a clear error message?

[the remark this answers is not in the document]
ichung: done

PR #350, opened 2019-01-22
build system: port to bazel

### LINKED ISSUE #347: Upgrade to Bazel
We've migrated through multiple build systems, each far more usable than the last, but Glass is not cutting it, and we need something better -- and preferably something we don't have to develop ourselves.

So we're going with [Bazel](https://bazel.build/).

celskeggs:
This PR involves a port of the entire build system from Glass to Bazel. It is split up into individual commits that should walk you through the different parts to this. Intermediate commits should not be expected to be functional.

The only reason this patch is so big is because it includes a 61,276-line autogenerated kubernetes patch. It would be +5,914 / -7,287 otherwise, for a net removal of ~1400 lines.

Fixes #347.

---

Checklist:


ichung:
:fire: :fire: :fire:

krawthekrow:
Not critical, but it looks like zip_file.bzl (commit e0f3640 in the PR) provides an example for how you can define a rule that runs Python directly instead of bashing out, and would probably be a cleaner way to do template substitution.

celskeggs:
Oh, yeah, it's absolutely possible to do the template substitution in starlark. I just haven't done the work to implement it like that yet.

platform/auth-monitor/auth-monitor.go
```diff
@@ -11,16 +11,16 @@ import (
 	"crypto/tls"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sipb/homeworld/platform/keysystem/api"
+	"github.com/sipb/homeworld/platform/keysystem/api/reqtarget"
+	"github.com/sipb/homeworld/platform/keysystem/api/server"
+	"github.com/sipb/homeworld/platform/util/osutil"
```
ichung: all go files: should organize commits into sections for clarity

celskeggs: Will resolve this in a follow-up PR, for the sake of getting the core changes merged ASAP.

celskeggs:
@krawthekrow can you re-review? I only added two commits.

celskeggs:
Aww squid. It's broken because of the debian ISO issue. I'll fix this later.

PR #352, opened 2019-02-07
[WIP] Add HTTPS support for apt

krawthekrow:
Jenkins is broken, probably because I tried to enable HSTS. Simply disabling HSTS didn't fix it, so I'm seeing if patching it from the other end would work. This is done by adding HTTPS support for apt.

I'm also piggybacking some documentation that should have been added with the Jenkins PR.

---

Checklist:

 - [ ] I have built my changes locally, and tested that the changes work as expected.
 - [ ] I have deployed a cluster incorporating my changes, and checked that it deploys successfully.

### DIFF
--- a/docs/jenkins-setup.txt
+++ b/docs/jenkins-setup.txt
@@ -1,0 +1,180 @@
+=== Machine setup ===
+- Download Debian netinst, check sha512sum (and signature of sha512sum), use Startup Disk Creator (preinstalled on Ubuntu) to make flash drive
+- In BIOS (F2, and then F2 again later), set flash drive mode to hard disk and put flash drive to the top of boot sequence
+- Load firmware from http://cdimage.debian.org/cdimage/unofficial/non-free/firmware/stretch/current/, need bnx2 and qlogic
+	- Via usb:
+		lsblk
+		mkdir /media/<drive name> # <drive name> is entirely arbitrary
+		mount /dev/sdb1 /media/<drive name>
+		(remember to umount afterwards)
+- Network config:
+	- address: 18.4.60.253
+		(from `stella hijinks`)
+	- netmask: 255.255.254.0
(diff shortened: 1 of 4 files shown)


docs/jenkins-setup.txt
```diff
@@ -0,0 +1,161 @@
...
+			for network gateways)
+	- nameservers: 18.72.0.3 18.70.0.160 18.71.0.151
+		(from `dig mit.edu` from within MIT's network)
+
+=== Standard setup ===
+apt update
+apt upgrade
+apt dist-upgrade
+apt install sudo openssh-server git
+apt install vim htop less xclip
```
celskeggs: Why do we need xclip on a (presumably headless) server?

docs/jenkins-setup.txt
```diff
@@ -0,0 +1,161 @@
...
+	- Pipeline: Stage View
+	- Git
+	- Github
+# create user
+# jenkins url: https://hijinks.mit.edu/jenkins/
+systemctl restart jenkins # jenkins bug
+
+=== Jenkins Homeworld ===
+# Manage Jenkins > Configure System > set # of executors 1
+	(protection in case locks don't work)
```
celskeggs: Do locks work? (Can we change this?)

docs/jenkins-setup.txt
krawthekrow: It's used to share clipboards over SSH in vim.

docs/jenkins-setup.txt
krawthekrow: Locks should work, but there currently isn't any reason to have more than one executor, so I just left the setting as is.

I could reword/remove the comment though.

krawthekrow:
I reworded the comment and also added the steps related to setting up Kerberos authentication.

PR #356, opened 2019-02-22
Support incremental upload versioning

celskeggs:
Currently, our bazel support only allows bumping the versions of all packages at once. This is problematic, because we need to reupload more than half a gigabyte of binaries every time we change anything, in order to make sure that an apt upgrade will actually pick up the changes.

---

Checklist:


platform/bazel/package.bzl
```diff
@@ -1,5 +1,5 @@
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar", "pkg_deb")
-load("//bazel:version.txt", "hash_compute")
+load("//bazel:version.bzl", "hash_compute", "version_compute")
```
krawthekrow: The version.txt -> version.bzl change should be in the hash_compute commit, if I understand correctly.

PR #382, opened 2019-04-19
Persistent autodeploy

krawthekrow:
This adds a "--persistent" flag to the cluster autodeploy so that we can easily mess around with a correctly autodeployed cluster. Note that because of how argparse arguments are matched to optional function parameters, this means that no spire function should have "params", "invoke" or "parser" as optional parameters, and seq functions can't have "dry_run" or "dry_run_outer" (we can improve this by renaming those things to things we will never use).

(no issue assigned).

---

Checklist:

### DIFF
--- a/platform/spire/src/command.py
+++ b/platform/spire/src/command.py
@@ -34,7 +34,7 @@ def get_command_for_function(f):
 
 def mux_map(desc: str, mapping: dict):
     def configure(command: list, parser: argparse.ArgumentParser):
-        parser.set_defaults(parser=parser)
+        parser.set_defaults(argparse_parser=parser)
         subparsers = parser.add_subparsers()
 
         for component, (inner_desc, inner_configure) in mapping.items():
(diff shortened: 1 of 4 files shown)


platform/spire/src/command.py
```diff
@@ -75,7 +75,10 @@ def invoke(args):
             fail("not enough parameters (expected %s)" % expect)
         if maxarg is not None and len(params) > maxarg:
             fail("too many parameters (expected %s)" % expect)
-        func(*params)
+        varnames = func.__code__.co_varnames
+        opts = vars(args)
```
celskeggs: Are you sure this is the recommended way to do this? It seems like it would have weird edge case things, like maybe including "params" as an option, when it's really the list of parameters. Is there no easier way to do this?

krawthekrow: I couldn't find an easier way to do this. If it helps, as suggested, I could rename "params", "invoke", etc. so that they will be unlikely to conflict with actual function parameters.

celskeggs: Apparently `vars` _is_ the recommended way to do this. Huh.

Issue #385, opened 2019-04-22
Ask admins to check tokens instead of asking admins to enter tokens


celskeggs:
It would be really nice if we could have a lower-overhead install process, where a token didn't need to be entered during the install phase.

A nicer way:
 - Have the node request a token from the supervisor
 - Have the node display it on the boot screen
 - Have the admin request a list of pending admissions from the supervisor
 - Have the admin confirm particular entries from the list of admissions

celskeggs:
Replaced/elaborated on by #426.

PR #387, opened 2019-04-23
Install non-supervisor nodes in parallel

krawthekrow:
Currently, we install the two non-supervisor nodes one by one during autodeploy. Installing both in parallel is a quick way to make the process significantly faster.

(no issue assigned)

---

Checklist:

 - [ ] I have deployed a cluster incorporating my changes, and checked that it deploys successfully. **verified by CI**

### DIFF
--- a/platform/spire/src/virt.py
+++ b/platform/spire/src/virt.py
@@ -1,5 +1,6 @@
 import argparse
 import atexit
+import concurrent.futures
 import contextlib
 import hashlib
 import os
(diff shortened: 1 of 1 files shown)


krawthekrow:
This reduces Jenkins autodeploy time by about 8 minutes, expect a similar speedup after #385. This is an autodeploy-only change, no manual deploy should be necessary.

cryslith:
Will this clean itself up properly if there's an error?  `concurrent.futures` might do a better job of unwinding our cleanup stack (and is also generally nicer).

krawthekrow:
@cryslith I've converted it to `concurrent.futures`. Any exceptions should now propagate to the master thread.

platform/spire/src/virt.py
```diff
@@ -491,6 +492,21 @@ def auto_node(ops: setup.Operations, tc: TerminationContext, node: configuration
     ops.add_operation("start up node @HOST", lambda: vm.boot_launch(), node)
```
celskeggs: If we're no longer using auto_node anywhere, we should get rid of it.

platform/spire/src/virt.py
```diff
@@ -491,6 +492,21 @@ def auto_node(ops: setup.Operations, tc: TerminationContext, node: configuration
...
 
 
+def auto_nodes_parallel(ops: setup.Operations, tc: TerminationContext, nodes: list, install_iso: str):
+    vms = [VirtualMachine(node, tc, install_iso) for node in nodes]
+
+    def boot_install_and_admit_all():
+        with concurrent.futures.ThreadPoolExecutor(len(vms)) as executor:
+            futures = [executor.submit(vm.boot_install_and_admit) for vm in vms]
+            results = [f.result() for f in futures]
```
celskeggs: Do we use `results` anywhere?

platform/spire/src/virt.py
```diff
@@ -491,6 +492,21 @@ def auto_node(ops: setup.Operations, tc: TerminationContext, node: configuration
...
+
+    ops.add_operation("install non-supervisor nodes (this may take several minutes)", boot_install_and_admit_all)
+
+    for node in nodes:
+        vm = VirtualMachine(node, tc)
+        ops.add_operation("start up node @HOST", vm.boot_launch, node)
```
celskeggs: Can we also parallelize launching the nodes?

celskeggs:
Also, needs rebase.

platform/spire/src/virt.py
krawthekrow: Does launching the nodes take a nontrivial amount of time? I have the impression it just starts the process and returns, since the nodes would be kept up for verification.

celskeggs: Oh, woops, yeah, you're right.

Issue #371, opened 2019-04-06
Make it easier to authenticate to the cluster using the disaster recovery key


cryslith:
Especially during testing it's usually more convenient to use the disaster recovery key than other forms of authentication.  Currently the only way to do this is to remove all entries from the `root-admins` section of `setup.yaml`, which is unintuitive and unnecessarily inconvenient.  We should change spire so that it tries the disaster recovery key if other methods of authentication fail.  It should also inform the user of what authentication method it used.

celskeggs:
Resolved by #396.

Issue #307, opened 2018-07-24
Improve IP class in spire


celskeggs:
> I notice there's some suspect things about [...] the IP class:
>
>    implementation of __eq__, __ne__ are incorrect because objects of different types should compare inequal
>    implementation of __and__ should raise an error instead of returning one
>    The whole class could probably be replaced with uses of the ipaddress standard library module

[Source](https://github.com/sipb/homeworld/pull/304#issuecomment-405993387)

Issue #303, opened 2018-07-17
asserts used as data validation


cryslith:
There are several places where python `assert` statements appear to be used for data validation, where a test-and-raise would be more appropriate.  It is not guaranteed by the semantics of `assert` that if the condition is not met, an exception is raised.  Unless we expect that we'll want to optimize the assertions away for efficiency, I think we should just change them all to test-and-raise.

celskeggs:
This seems like a good idea. Feel free to assign yourself if you think this should be done now; otherwise I'll put it on the backburner and take care of it later.

celskeggs:
@cryslith do you consider this complete after #396's merger?

cryslith:
yesOn Nov 14, 2019 10:40, Cel Skeggs <notifications@github.com> wrote:@cryslith do you consider this complete after #396's merger?

Issue #372, opened 2019-04-06
Handle when root-admins is None instead of []


krawthekrow:
Or add some better documentation about Kerberos-less deploys.

celskeggs:
Is this better solved by #371?

krawthekrow:
Yes.

PR #396, opened 2019-05-05
spire improvements

cryslith:
Rewrite most of the spire `command`, `Operations` infrastructure to be easier to change.  Also add a couple features along the way.

See #371 #307 #303

---

Checklist:


platform/spire/src/access.py
```diff
@@ -86,13 +89,21 @@ def renew_ssh_cert() -> str:
...
+        if configuration.get_config().is_kerberos_enabled():
+            print("rotating", variant, "certs via keyreq")
+            if ca_path is None:
+                call_keyreq(variant + "-cert", key_path, cert_path)
+            else:
+                call_keyreq(variant + "-cert", key_path, cert_path, ca_path)
+            return
+    except Exception as e:
+        print("[keyreq failed]")
+        errs.append(e)
```
celskeggs: This appears to cause errors to not be explained if keyreq fails but cert bypass succeeds, which would make it hard to debug a keyreq failure, because no details would be outputted besides `[keyreq failed]`. Is there any way that this can be easier? (Like a `--debug` or `--explain` option or something?)

platform/spire/src/command.py
```diff
@@ -27,6 +27,14 @@ def fail(message: str, hint: str = None) -> None:
...
 
 
+class MultipleExceptions(Exception):
+    def __init__(self, message, errs):
+        self.errs = errs
+        super().__init__('\n'.join([message] + [str(e) for e in errs]))
+        self
+        messages = [message] + [str(e) for e in errs]
+        hints = (e.hint for e in errs if hasattr(e, 'hint'))
+        super().__init__('\n'.join(messages), '\n'.join(hints))
```
celskeggs: We already talked about this, but just to have a record: this looks like it's spliced wrong.

platform/spire/src/infra.py
```diff
@@ -8,11 +8,24 @@
...
-        return access.call_keyreq("bootstrap-token", principal_hostname).decode().strip()
-    else:
+
+    errs = []
+
+    try:
+        if config.is_kerberos_enabled():
+            return access.call_keyreq("bootstrap-token", principal_hostname).decode().strip()
+    except Exception as e:
+        errs.append(e)
```
celskeggs: This has the same problem as before, where errors are hidden, but it doesn't look like this even notifies the user that it had to resort to keyinitadmit instead of keyreq.

platform/spire/src/iso.py
```diff
@@ -108,9 +108,10 @@ def gen_iso(iso_image, authorized_key, mode=None):
         inclusion += ["sshd_config.new", "preseed.cfg"]
 
         for package_name, (short_filename, package_bytes) in packages.verified_download_full(PACKAGES).items():
-            assert "/" not in short_filename, "invalid package name: %s for %s" % (short_filename, package_name)
-            assert short_filename.startswith(package_name + "_"), "invalid package name: %s for %s" % (short_filename, package_name)
-            assert short_filename.endswith("_amd64.deb"), "invalid package name: %s for %s" % (short_filename, package_name)
+            if any(["/" in short_filename,
+                    not short_filename.startswith(package_name + "_"),
+                    not short_filename.endswith("_amd64.deb")]):
```
celskeggs: Why are we using `any([a, b, c])` instead of `a or b or c`?

cryslith:
For my reference:
- The "good" commits are `deploy-chroot: adjust spire-reinstallation scripts`, `spire: append -dirty to version`, and `spire: replace IP class with ipaddress standard library`
- The long commit that needs splitting is `spire: refactor Command and Operations classes to be more extensible`

cryslith:
From Mattermost archives:
>> Also, what is the `# check that these work` comment for?
> I should rewrite that `# check that these work` comment, but what it's saying is that the call to `bind` doesn't actually do anything except throw an error early if the arguments are wrong somehow

>> Why does MultipleExceptions call its constructor twice?
> I think I made a typo in MultipleExceptions somehow while splitting commits; will fix

cryslith:
Need to update docs to reflect new usage of deploy-chroot reinstall scripts

cryslith:
need to remove unused imports

cryslith:
@celskeggs I still need to test this out, but can you review for now?  Hopefully the commits are much more readable this time

platform/spire/src/access.py
```diff
@@ -42,7 +42,7 @@ def needs_rotate(path, interval=DEFAULT_ROTATE_INTERVAL):
```
celskeggs: This commit message overflows the maximum allowable length of the first line, according to GitHub.

platform/spire/src/access.py
```diff
@@ -87,13 +90,23 @@ def renew_ssh_cert() -> str:
...
+            print("rotating", variant, "certs via keyreq")
+            if ca_path is None:
+                call_keyreq(variant + "-cert", key_path, cert_path)
+            else:
+                call_keyreq(variant + "-cert", key_path, cert_path, ca_path)
+            return
+    except Exception as e:
+        print("[keyreq failed]")
+        if os.environ.get('SPIRE_DEBUG'):
+            traceback.print_exc()
```
celskeggs: Could we have `[keyreq failed]` include a note that you can set SPIRE_DEBUG to get more info?

platform/spire/src/command.py
```diff
@@ -27,6 +27,13 @@ def fail(message: str, hint: str = None) -> None:
     raise CommandFailedException(message, hint)
 
 
+class MultipleExceptions(CommandFailedException):
+    def __init__(self, message, errs):
+        messages = [message] + [str(e) for e in errs]
+        hints = (e.hint for e in errs if hasattr(e, 'hint'))
+        super().__init__('\n'.join(messages), '\n'.join(hints))
```
celskeggs: Is this guaranteed to be sensical? i.e. if the output message ends up having multiple lines and it goes:
```
[message]
[message for error 1]
[message for error 2]
[hint for error 1]
[hint for error 2]
```
It seems like it might be confusing which hint goes with which error.

platform/spire/src/virt.py
```diff
@@ -19,18 +19,18 @@
...
 
 def get_bridge(ip):
-    return "spirebr%s" % hex(ip.to_integer())[2:].upper()
+    return "spirebr%s" % ip.packed.hex().upper()
 
 
 def get_node_tap(node):
     # maximum length: 15 characters
-    return "spirtap%s" % hex(node.ip.to_integer())[2:].upper()
+    return "spirtap%s" % node.ip.packed.hex().upper()
```
celskeggs: Did we determine that this was going to change the format slightly, due to padding? I seem to remember we hit some weird issues around `spire virt net down` as a result of this, so it might be worth calling out in the commit message.

platform/spire/src/configuration.py
```diff
@@ -1,3 +1,4 @@
+from ipaddress import IPv4Address, IPv4Network
```
celskeggs: w/r/t the commit message: I've been trying to avoid putting actual issue #'s in my commit messages, because it wreaks havoc on GitHub's mention tracking when you keep rebasing commits. (See #307's history for what I'm talking about.)

deploy-chroot/reinstall-spire.sh
```diff
@@ -1,14 +1,6 @@
...
-    echo '(have you created a chroot?)'
-    exit 1
-fi
-
-cd "$(dirname "$0")"
-cd "$(git rev-parse --show-toplevel)"
-sudo systemd-nspawn -E PATH="/usr/local/bin:/usr/bin:/bin" -M "$(basename $HOMEWORLD_DEPLOY_CHROOT)" --bind $(pwd):/homeworld:norbind -u "$USER" -a -D "$HOMEWORLD_DEPLOY_CHROOT" bash -c "cd /homeworld && (sudo apt-get -y purge homeworld-apt-setup && sudo apt-get -y remove 'homeworld-*' && sudo apt-get clean || true) && sudo dpkg -i apt-setup.deb && sudo apt-get update && sudo apt-get -y install homeworld-spire"
+sudo apt-get -y remove 'homeworld-spire' && sudo apt-get clean || true
+sudo apt-get update
+sudo apt-get -y install homeworld-spire
```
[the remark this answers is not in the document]
cryslith: I think this was before we had automatically-incrementing debian package numbers, so doing this was easier for development than incrementing the version.  Shouldn't be necessary anymore

docs/cluster-redeploy.md
```diff
@@ -0,0 +1,144 @@
...
+
+    $ scp preseeded.iso toast:/srv/preseeded.iso
+    $ ssh toast
+    toastfs-dev:# hyades/rebuild-homeworld-cluster.sh /srv/preseeded.iso
+
+You can then access the virtual machines using VNC.
+For example, using TigerVNC:
+
+    $ sudo apt-get install tigervnc
+    $ ssh toast-vnc & # leave this running in the background
```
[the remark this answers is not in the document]
cryslith: not quite, but there is a version that does work: `ssh -n toast-vnc 'sleep infinity'`

platform/spire/src/command.py
```diff
@@ -107,15 +117,9 @@ def main_invoke(command):
...
-        print(ANSI_ESCAPE_CODE_RED + 'command failed: ' + str(e) + ANSI_ESCAPE_CODE_RESET)
-        if e.hint is not None:
-            print(ANSI_ESCAPE_CODE_YELLOW + e.hint + ANSI_ESCAPE_CODE_RESET)
-        return 1
+    args = parser.parse_args()
+    if "argparse_invoke" in args:
+        args.argparse_invoke(args)
+    else:
+        args.argparse_parser.print_help()
+    return 0
```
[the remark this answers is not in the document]
cryslith: The overall point of this commit is actually so that `MultipleExceptions` can later inherit the nice colored messages from `CommandFailedException`.

platform/spire/src/command.py
```diff
@@ -27,16 +27,23 @@ def fail(message: str, hint: str = None) -> None:
     raise CommandFailedException(message, hint)
 
 
-def mux_map(desc: str, mapping: dict):
-    def configure(command: list, parser: argparse.ArgumentParser):
+class Mux:
+    def __init__(self, description, mapping):
+        self.__doc__ = description
+        self.mapping = mapping
```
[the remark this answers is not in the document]
cryslith: We use the `__doc__` property of function commands in help messages.  I chose to propagate that property into `Command`, which means that `__doc__` is part of the interface which commands are supposed to fulfill.  `Mux` is a type of command, so it needs to fulfill the same interface.  We could use a different property, I don't have a strong opinion.

platform/spire/src/command.py
```diff
@@ -27,16 +27,23 @@ def fail(message: str, hint: str = None) -> None:
...
-            inner_configure(command + [component], inner_parser)
+        for component, subcommand in self.mapping.items():
+            subparser = subparsers.add_parser(component, description=subcommand.__doc__, help=subcommand.__doc__)
+            try:
+                subcommand.configure(command + [component], subparser)
+            except AttributeError as e:
+                raise Exception("error configuring subcommand {!r}".format(subcommand)) from e
 
-    return desc, configure
+mux_map = Mux
```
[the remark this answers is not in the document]
cryslith: I think at some point I had the idea they would be a drop-in replacement for the old versions, but yeah that isn't the case

platform/spire/src/command.py
```diff
@@ -162,6 +162,85 @@ def wrap(f):
     return functools.update_wrapper(Simple(f), f, updated=[])
 
 
+def _delegate_to_context(f):
+    @functools.wraps(f)
+    def g(self, *args, **kwargs):
+        if self._context is None:
+            return f(self, *args, **kwargs)
+        return f(self._context, *args, **kwargs)
+    return g
```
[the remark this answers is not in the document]
cryslith: Yes if I can remember why :cold_sweat:

platform/spire/src/access.py
cryslith: github comments are associated with files, not commits.  but I will go back and check all of them

platform/spire/src/command.py
cryslith: good question, will fix

platform/spire/src/configuration.py
cryslith: oh no

platform/spire/src/access.py
celskeggs: oh no woops. I'm talking about d657905337588ed7dbbbdaeb983eb1d1fa08aee4.

platform/spire/src/command.py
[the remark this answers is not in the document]
celskeggs: oh, makes sense! (I'm sure this is a redundant request at this point, but can we add that to the commit message?)

platform/spire/src/command.py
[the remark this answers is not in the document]
celskeggs: Are the function commands and Command supposed to be compatible, and is that why `__doc__` is named that? If it's just a name, and isn't needed to be this for compatibility, then I do think I'd encourage using another name.

platform/spire/src/command.py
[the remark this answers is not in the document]
celskeggs: mood tbh

platform/spire/src/command.py
```diff
@@ -54,69 +39,125 @@ def configure(command: list, parser: argparse.ArgumentParser):
...
-        for _, p in sig.parameters.items():
-            # TODO: make ops an optional argument in all wrapped functions
-            # instead of the first required argument
-            if p.name == "ops":
+        for name, p in self.sig.parameters.items():
+            if p.kind == inspect.Parameter.POSITIONAL_OR_KEYWORD:
+                if p.default == inspect.Parameter.empty:
+                    parser.add_argument(name)
+                    continue
+                if p.annotation == bool and not p.default:
```
cryslith: note: need to be stricter about unintended ways to set arguments

platform/spire/src/command.py
```diff
@@ -54,69 +39,125 @@ def configure(command: list, parser: argparse.ArgumentParser):
...
+    return functools.update_wrapper(Command(f), f, updated=[])
+
+
+class Simple(Command):
+    def remove_ops_from_sig(self):
+        pass
+
+    def operate(self, op, *args, **kwargs):
+        op.add_operation(self.__doc__, lambda: self.func(*args, **kwargs),
+                         self.command(*args, **kwargs))
```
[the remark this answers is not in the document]
cryslith: `__doc__` does actually get set in this commit, it's passed through by `functools.update_wrapper`

platform/spire/src/command.py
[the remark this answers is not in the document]
cryslith: Looking back at how this is structured, I believe that yes they need to be compatible, because the `Command` replaces the original function.

platform/spire/src/setup.py
```diff
@@ -91,28 +91,31 @@ def run_operations(self) -> None:
...
-    def ssh_upload_bytes(self, name: str, node: configuration.Node, source_bytes: bytes, dest_path: str) -> None:
-        self.add_operation(name, lambda: ssh.upload_bytes(node, source_bytes, dest_path), node=node)
+def ssh_raw(ops, name: str, node: configuration.Node, script: str, in_directory: str=None, redirect_to: str=None)\
+        -> None:
+    if redirect_to:
+        script = "(%s) >%s" % (script, escape_shell(redirect_to))
+    if in_directory:
+        script = "cd %s && %s" % (escape_shell(in_directory), script)
+    ops.add_operation(name.replace('@HOST', node.hostname),
+                      lambda: ssh.check_ssh(node, script))
```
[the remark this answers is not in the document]
cryslith: Well, I would call this behavior-preserving.  The [usage](https://github.com/sipb/homeworld/pull/396/commits/93062723afcf060b888a1dc37a309eaf51b2daa1#diff-56e3b0e32537d3e7b77086d6058ef3e9L232-R247) is the same (except that `ops` is the first argument rather than the receiver).

docs/cluster-autodeploy.md
```diff
@@ -0,0 +1,24 @@
...
+Here `<x>` is a byte which must be unique per machine
+if you plan to do multiple concurrent autodeploys.
+If you are working on rhombi, please coordinate assignment of these numbers
+via the `/lock` bot on mattermost.
+
+## Autodeploy
+
+    $ spire virt auto install
+
+That's it.
```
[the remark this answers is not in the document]
cryslith: Why is it useful to run screen?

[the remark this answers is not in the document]
celskeggs: So that your autodeploy doesn't get aborted if your SSH session disconnects?

deploy-chroot/reinstall-apt-setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e -u
+
+sudo apt-get -y purge homeworld-apt-setup && sudo apt-get -y remove 'homeworld-*' && sudo apt-get clean || true
```
celskeggs: What is the purpose of the '|| true' here? Wouldn't we want to know if one of these three steps were to fail?

docs/cluster-autodeploy.md
```diff
@@ -0,0 +1,22 @@
...
+    $ deploy-chroot/generate-setup.py <x>
+
+Here `<x>` is a byte which must be unique per machine
+if you plan to do multiple concurrent autodeploys.
+If you are working on rhombi, please coordinate assignment of these numbers
+via the `/lock` bot on mattermost.
+
+## Autodeploy
+
+    $ eval `ssh-agent -s` && ssh-add  # if not already running
```
celskeggs: Might want to clarify what "if not already running" means for people who don't realize this will start a SSH agent?

deploy-chroot/reinstall-apt-setup.sh
cryslith: I do not remember why that is there

docs/cluster-autodeploy.md
cryslith: There is a more verbose explanation in the other doc, and I don't feel like explaining basic things every time in these documents.

docs/cluster-autodeploy.md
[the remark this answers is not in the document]
cryslith: That seems orthogonal to this doc.  Among other things this might not even be run over SSH.

deploy-chroot/reinstall-apt-setup.sh
cryslith: Oh, it's so that this works the first time you run the script.

celskeggs: Do we need it on the `apt-get clean` command?

platform/spire/src/__main__.py
```diff
@@ -15,7 +15,7 @@
 import version
 import virt
 
-main_command = command.mux_map("invoke a top-level command", {
+main_command = command.Mux("invoke a top-level command", {
```
celskeggs: There are a bunch of the commits between the method signature changes and this commit that aren't marked as being expected not to function, even though it seems like they wouldn't function due to the signature mismatches not yet being resolved.

I'm worried about that, because I could easily imagine landing on one of those during bisecting, and assuming that the commit works correctly. Could we mark those intermediate commits as not being expected to function?

platform/spire/src/configuration.py
```diff
@@ -179,6 +179,7 @@ def __init__(self, kv: dict):
                     command.fail("in config: multiple supervisors not yet supported")
                 self.keyserver = node
 
+    # TODO(#372): make this configuration setting more explicit
```
celskeggs: Do you mean #371? #372 is closed.

platform/spire/src/infra.py
```diff
@@ -1,6 +1,7 @@
 import access
 import command
 import configuration
+import os
```
celskeggs: Should adding this import go in the commit where the relevant functionality was added?

deploy-chroot/reinstall-apt-setup.sh
cryslith: If the `apt-get remove` fails then we don't want to run `apt-get clean` at all.

platform/spire/src/command.py
```diff
@@ -54,69 +39,136 @@ def configure(command: list, parser: argparse.ArgumentParser):
...
+        return self.func(*args, **kwargs)
+
+    def operate(self, op, *args, **kwargs):
+        "Schedule this command to be run by Operations"
+        self.func(op, *args, **kwargs)
+
+    def process_args(self, cli_args):
+        "Process command-line arguments into function arguments"
+        # cli_args is the arguments object produced by argparse
+        cli_args = vars(cli_args)
```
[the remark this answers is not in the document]
cryslith: I think I just have a deeply held, unique and fundamentally incompatible stance on this issue from other people, since it just came up at work too.  _sigh_.  Yes, I will change the variable name :stuck_out_tongue:

platform/spire/src/command.py
```diff
@@ -177,6 +177,108 @@ def wrap(f):
...
+        >>> with ctx:
+        ...     some_function()
+
+        were written.
+
+        For convenience,
+        the original context ctx is made available to the context body as c.
+        """
+
+        # TODO: figure out annotations so that this can be processed correctly by --dry-run
```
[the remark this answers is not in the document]
cryslith: I think it is not accomplished by this PR.

platform/spire/src/__main__.py
cryslith: Oops, I think the commits are in the wrong order somehow...

deploy-chroot/reinstall-apt-setup.sh
celskeggs: Hmm, okay.

platform/spire/src/command.py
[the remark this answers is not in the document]
celskeggs: Okay, cool.

platform/spire/src/__main__.py
cryslith: Actually - what is an example of such a commit?  I don't see any

platform/spire/src/infra.py
cryslith: This import should go in `spire: attempt multiple authentication methods` (but this would require a rebase).

platform/spire/src/__main__.py
celskeggs: There was 7e8719ce4244ae8b4b46f300f3f3dc84109360a1, but that was covered by another comment I made. I certainly can't see the same thing I remember seeing anymore, which means either some sort of weird Git thing happened or my eyes are fake. In any case, this looks good in the latest rebase you've done.

cryslith: I think at least one of GitHub's views does not display the commits in order

celskeggs: I went through via the Files Changed commit iteration thing, which I think is in topological order?

cryslith: :woman_shrugging: :crab:

Issue #256, opened 2018-02-12
Fix kubectl logs


celskeggs:
At least one issue is that kubectl assumes that it can directly resolve the node names, rather than needing to convert them to a FQDN.

In any case, we should debug until we can get `kubectl logs` to work.

celskeggs:
Yes, the problem IS that the node name is accessed directly, instead of the FQDN. Working around this by adding entries to /etc/hosts allows the logs to be fetched successfully.

celskeggs:
@krawthekrow I'm under the impression that you fixed this in #399; please let me know if this was not the case.

PR #399, opened 2019-05-20
Fix kubelet hostname

krawthekrow:
See commit message.

See #256.

---

Checklist:

### DIFF
--- a/platform/spire/resources/setup.yaml
+++ b/platform/spire/resources/setup.yaml
@@ -18,8 +18,7 @@ dns-upstreams:
   - 18.71.0.151
   - 18.72.0.3
 
-dns-bootstrap:
-  homeworld.private: <supervisor node>
+dns-bootstrap: {}
 
 root-admins:
   - example/root@ATHENA.MIT.EDU
--- a/platform/spire/src/setup.py
+++ b/platform/spire/src/setup.py
@@ -226,7 +226,17 @@ def modify_dns_bootstrap(ops: Operations, is_install: bool) -> None:
 
 def dns_bootstrap_lines() -> str:
     config = configuration.get_config()
-    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (ip, hostname) for hostname, ip in config.dns_bootstrap.items())
+    dns_hosts = config.dns_bootstrap.copy()
+    dns_hosts["homeworld.private"] = config.keyserver.ip
+    for node in config.nodes:
+        full_hostname = "%s.%s" % (node.hostname, config.external_domain)
+        if node.hostname in dns_hosts:
+            command.fail("redundant /etc/hosts entry: %s", node.hostname)
+        if full_hostname in dns_hosts:
+            command.fail("redundant /etc/hosts entry: %s", full_hostname)
+        dns_hosts[node.hostname] = node.ip
+        dns_hosts[full_hostname] = node.ip
+    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (ip, hostname) for hostname, ip in dns_hosts.items())
 
 
 def setup_dns_bootstrap(ops: Operations) -> None:
--- a/.jenkins/deploy-setup.yaml
+++ b/.jenkins/deploy-setup.yaml
@@ -18,11 +18,7 @@ dns-upstreams:
   - 18.71.0.151
   - 18.72.0.3
 
-dns-bootstrap:
-  homeworld.private: 10.1.0.2
-  jet.pink.example.org: 10.1.0.2
-  jade.pink.example.org: 10.1.0.3
-  jasper.pink.example.org: 10.1.0.4
+dns-bootstrap: {}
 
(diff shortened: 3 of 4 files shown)


celskeggs:
@krawthekrow how's this going?

krawthekrow:
I had decided that adding the non-fully-qualified hostname to hosts was the better option, since override-hostname seemed to create other problems and may be deprecated. Unfortunately my local test from last week failed because of the ISO regression, and then I got very hosed.

I'll give it another poke.

krawthekrow:
Note: it doesn't look like there's a way to update /etc/hosts by modifying setup.yaml when the cluster is live.

platform/spire/src/setup.py
```diff
@@ -228,7 +228,9 @@ def modify_dns_bootstrap(ops: Operations, is_install: bool) -> None:
 
 def dns_bootstrap_lines() -> str:
     config = configuration.get_config()
-    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (ip, hostname) for hostname, ip in config.dns_bootstrap.items())
+    lines = [(ip, hostname) for hostname, ip in config.dns_bootstrap.items()]
+    lines += [(node.ip, node.hostname) for node in config.nodes]
+    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (line[0], line[1]) for line in lines)
```
celskeggs: Couple optional style comments:
 - Might be clearer if you use a more specific variable name than `lines`
 - Might be clearer to do ` for ip, hostname in lines` than ` for line in lines` and have to index into each line.

celskeggs:
Let me know if you want to address the above optional comments or if you just want this to be merged.

celskeggs:
On further consideration, I'm unsure whether this is the right solution -- it seems weird to specify _some_ hostnames for bootstrapping in setup.yaml's `dns-bootstrap` section, and have some be implicitly added.

Some possible solutions:
 - Maybe this should be based on stripping the suffix off of fqdns specified in the `dns-bootstrap` section
 - Maybe we should be autogenerating those parts of the `dns-bootstrap` section too in this code

What do you think?

krawthekrow:
Yeah, I thought it was strange too that the dns-bootstrap section wasn't autogenerated. If there was no good reason behind that, I don't see why we shouldn't.

I don't see a use case where those lines _shouldn't_ be added, anyway.

celskeggs:
The reason they were like that is that we intended for it to be an interstitial state, not the way the cluster always operates: the concept being that we'd be self-hosting DNS for our cluster, so once it was up, we could use our own DNS server, but until then, we'd have to bootstrap by directly populating the hosts file.

But I think that'll require enough additional work that we needn't try to support that too much right now. I think generating the rest of the entries too is a good idea for now.

platform/spire/src/setup.py
```diff
@@ -226,7 +226,13 @@ def modify_dns_bootstrap(ops: Operations, is_install: bool) -> None:
 
 def dns_bootstrap_lines() -> str:
     config = configuration.get_config()
-    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (ip, hostname) for hostname, ip in config.dns_bootstrap.items())
+    dns_hosts = list(config.dns_bootstrap.items())
+    for node in config.nodes:
+        if node.kind == "supervisor":
+            dns_hosts += [("homeworld.private", node.ip)]
```
celskeggs: This won't work correctly if we have more than one supervisor. That's fine, since we already don't support having more than one supervisor, but can we make the code fail fast in this situation, instead of silently generating a bad `/etc/hosts` file?

One way to do this would be to just use the `config.supervisor` property, which already fails-fast during generation when more than one supervisor is specified; that way, if we ever support more than one, we'll notice that this code expects just one while repairing uses of that field.

platform/spire/resources/setup.yaml
```diff
@@ -18,8 +18,7 @@ dns-upstreams:
   - 18.71.0.151
   - 18.72.0.3
 
-dns-bootstrap:
-  homeworld.private: <supervisor node>
+dns-bootstrap: {}
```
celskeggs: Can you also update the configuration files in `deploy-chroot` and `.jenkins`?

platform/spire/src/setup.py
```diff
@@ -226,7 +226,13 @@ def modify_dns_bootstrap(ops: Operations, is_install: bool) -> None:
 
 def dns_bootstrap_lines() -> str:
     config = configuration.get_config()
-    return "".join("%s\t%s # AUTO-HOMEWORLD-BOOTSTRAP\n" % (ip, hostname) for hostname, ip in config.dns_bootstrap.items())
+    dns_hosts = list(config.dns_bootstrap.items())
```
celskeggs: Currently, it looks like we'll just generate a bad `/etc/hosts` file if the same hostname is specified in setup.yaml and generated here. Can we explicitly fail in that situation, or maybe explicitly override one or the other, if you'd prefer that behavior?

PR #406, opened 2019-09-03
bazel: pin bazel version in use

### LINKED ISSUE #378: Reproducibly select version of Bazel
Currently, we use the latest Bazel version in CI -- which means that sometimes master breaks because Bazel stopped supporting something one of our dependencies does.

It would be really nice to have builds that don't migrate from "working" to "broken" just because of an updated release getting published. We should find a way to install a particular version and bump it explicitly.

celskeggs:
Instead of using the latest available Bazel version in the upstream apt repo, pull down pinned Bazel versions from upstream GitHub, so that we can avoid having Bazel change out from under us.

Fixes #378.

---

Checklist:

 - [ ] ~I have deployed a cluster incorporating my changes, and checked that it deploys successfully.~


celskeggs:
Changes broke under Jenkins because curl wasn't installed outside of the build chroot. I've added curl to the package installation step.

celskeggs:
@cryslith can you re-review? I've added changes to jenkins-build.sh that were necessary for this to work under Jenkins.

cryslith:
Approved but I cannot press the button until I get home from workOn Sep 3, 2019 14:31, Cel Skeggs <notifications@github.com> wrote:@cryslith can you re-review? I've added changes to jenkins-build.sh that were necessary for this to work under Jenkins.

celskeggs:
Thanks! I'll just merge the thing directly.

PR #412, opened 2019-09-13
Refactor keyclient

### LINKED ISSUE #368: Simplify keysystem's configuration model
Currently, the configuration files for the keysystem are generated by a complicated set of pieces of code that perform things like textual substitution and swap out configuration files at various times.

To fully understand how the keysystem initializes, you need to understand spire, the keysystem code, the keysystem systemd scripts, and the preseeding process. It would be nice to make this easier to understand.

Really, the key problem AFAICT is that the contents of the keysystem configuration file is really more like a piece of code than a configuration file, and so we're essentially building a low-quality programming language. I suggest that we should just write the code in Go, and use the existing keysystem code as a library -- which would also radically simplify the configuration parsing code in the keysystem, making everything simpler.

celskeggs:
This is about half of the work needed to resolve issue #368. It involves refactoring the keyclient to no longer use a configuration file, and instead define the actions that should be taken inside the keyclient itself. It also involves other code cleanup and simplification, especially ones made possible by no longer having to parse configuration files.

This includes some changes to the keyserver, but is not the bulk of the keyserver refactoring work. There is likely to be even more room for refactoring once the keyserver is also refactored. This PR is being filed separately, because there's a lot more work left on the keyserver refactoring attempt, and I'd like to start getting this part reviewed.

The commits in this PR have been carefully written, and (if I did everything correctly) should each make sense as a single refactoring step in the keyclient. Each commit compiles and deploys successfully, or at least did last time I tried that particular commit.

I delete a bunch of unit tests as part of these changes; that is because the keysystem was overtested, in an attempt to make it usable for other uses. Unfortunately, it has become clear at this point that agility of the keysystem codebase is more important than test coverage, so it doesn't make sense to write as many new tests for the remaining functionality. We do now also have the automatic cluster deploy test, which is a more accurate integration test than anything we had.

---

Checklist:

 - [x] I have written or updated appropriate documentation to cover this change. (No documentation change is needed, because this change is not visible.)
 - [x] If my change includes new or modified functionality, I have tested that the changes work as expected. (No functionality changes are expected.)


cryslith:
Can you reword the PR message so that this won't automatically close #368?

celskeggs:
Done.

platform/keysystem/worldconfig/keyclient.go
```diff
@@ -51,145 +59,188 @@ func GetVariant() (string, error) {
...
+		Name:    "clustertls",
+		Path:    "/usr/local/share/ca-certificates/extra/cluster.tls.crt",
+		Refresh: "24h",
+		Mode:    "644",
+	})
+	b.Download(config.ConfigDownload{
+		Type:    "authority",
+		Name:    "ssh-user",
+		Path:    "/etc/ssh/ssh_user_ca.pub",
+		Refresh: "168h", // allow a week for mistakes to be noticed on this one
```
cryslith: OOC can you explain this comment?

platform/keysystem/worldconfig/keyclient.go
```diff
@@ -91,28 +91,28 @@ func (b *ActionBuilder) DefaultDownloads(variant string) {
...
+	b.PublicKey(
+		"etcd-client",
+		"/etc/homeworld/authorities/etcd-client.pem",
+		OneDay,
+	)
+	b.PublicKey(
+		"etcd-server",
+		"/etc/homeworld/authorities/etcd-server.pem",
+		OneDay,
+	)
```
cryslith: Does downloading these public keys implicitly trust them?  If so then this change is making these keys trusted in more places than they were before.  (Of course there's really not much difference between "master-only" and "all nodes" but I still think it's worth pointing out)

platform/keysystem/keyclient/state/state.go
```diff
@@ -15,6 +16,29 @@ import (
...
+
+func NewClientState(keyserver *server.Keyserver) *ClientState {
+	return &ClientState{
+		Keyserver: keyserver,
+		RetryAt:   make(map[string]time.Time),
+	}
+}
+
+// how frequently to recheck whether our access is denied, in case it changes
+const RetryAfter = time.Hour * 12
```
cryslith: Supposing we accidentally didn't grant enough permissions, is there an easy way we could manually force this to retry besides waiting 12 hours?  Would it suffice to restart whichever services were affected?

platform/keysystem/keyclient/actloop/actloop.go
```diff
@@ -20,45 +20,63 @@ type Action interface {
...
+			if blockerr != nil {
+				nac.Blocked(blockerr)
+			} else {
+				err = action.Perform(nac.Logger)
+				if err != nil {
+					nac.Errored(action.Info(), err)
+				} else {
+					nac.NotifyPerformed(action.Info())
+				}
+			}
```
cryslith: I have a strange feeling that we're implementing some weird version of promises.  Can we discuss the action loop design again?  It's been 1.5 years since the last time you explained to me why we need it.

platform/keysystem/worldconfig/keyclient.go
celskeggs: I don't remember exactly, but it's something like "ssh is very important and so we should update it less frequently, so that if things break badly, the ssh keys previously generated are more likely to still work when we realize things are broken"

platform/keysystem/worldconfig/keyclient.go
celskeggs: Only the keys in special directories like `/usr/local/share/ca-certificates/extra/`. Others are only pulled in by the particular services that need to validate them, which won't be running on the non-master nodes.

platform/keysystem/keyclient/state/state.go
celskeggs: Yes, retry timeouts aren't persisted to disk, so you can just roll the keyclient with `systemctl restart keyclient`. We already do this during the install process when we want it to pick up something immediately.

platform/keysystem/keyclient/actloop/actloop.go
celskeggs: The actloop is really just an abstraction to facilitate converging a bunch of pieces of state. Each action represents something that could need to be done, and (in the previous design) had functions to:
 - check whether the action needs to be done (Pending)
 - check whether there are any unsatisfied preconditions for the action (CheckBlocker)
 - attempt to perform the action (Perform)

The actloop keeps attempting to perform actions until there are no actions left that can be performed. Note that actions can go from a "done" state to a "needs to be done again" state, such as if a certificate is getting close to expiring, in which case it will be re-run. So I don't think it's quite like promises.

platform/keysystem/worldconfig/keyclient.go
celskeggs: I think the idea is that everything else has a shorter delay, so breakages would be likely to occur in non-SSH components first, whether in dev or prod, and we'd notice breakage through that first.
(In any case, this PR just preserves the existing behavior, so we should take this to a new issue if we think we should change it.)

celskeggs:
GitHub says that not all the checks passed on merge, but they had -- I just had to do a tiny rebase before merging, which was small enough (a single line) that I'm not considering it worth retesting for or re-asking for a review for.

PR #423, opened 2019-10-13
Implement RBAC on cluster

### LINKED ISSUE #12: Change apiserver to use ABAC or RBAC instead of AlwaysAllow
Kubernetes has two key phases of processing a request that are security-relevant:
 - Authentication: determining with confidence who or what is making this request.
 - Authorization: determining whether the entity making this request is allowed to make this particular request, or whether it's outside the scope of what they should be allowed to do.

For authentication, we're using a combination of kubernetes certificate authentication and serviceaccount-based authentication. This is likely to be something of a long-term solution.

For authorization, we're currently using AlwaysAllow -- which means that anyone authenticated to the cluster is allowed to perform any operation. This is insecure and a terrible idea.

This project is to review the available modes for kubernetes authorization (abac, rbac, webhooks), compare and contrast what they support, decide (with input from the homeworld primary maintainer) which one should be used, and then implement sample policy and/or policy generation for the chosen mode for use in the cluster. This also means testing the policy against the actual operations that will be performed and confirming that it doesn't break the cluster or allow privilege escalation that would compromise the cluster's security.

celskeggs:
This PR turns on Role-Based Access Control in the cluster. We already have authentication; now we add authorization. For more details about the set-up of the RBAC configuration, refer to the file added to design/ in the last commit. Beyond that, each commit should document its own details.

Resolves #12.

---

Checklist:


platform/kubernetes/launch.go
```diff
@@ -0,0 +1,269 @@
...
+		// public addresses
+		"--bind-address", "0.0.0.0", "--advertise-address", hostIP,
+		// IP range
+		"--service-cluster-ip-range", serviceCidr,
+		// use standard HTTPS port for secure port
+		"--secure-port", "443",
+		// etcd cluster to use
+		"--etcd-servers", etcdEndpoints,
+		// don't allow privileged containers: don't allow this kind of thing
+		"--allow-privileged", "true",
```
cryslith: what does this comment mean?

platform/kubernetes/launch.go
```diff
@@ -0,0 +1,269 @@
...
+		// authenticate the etcd cluster to us
+		"--etcd-cafile", "/etc/homeworld/authorities/etcd-server.pem",
+		// authenticate us to the etcd cluster
+		"--etcd-certfile", "/etc/homeworld/keys/etcd-client.pem", "--etcd-keyfile", "/etc/homeworld/keys/etcd-client.key",
+		// disallow insecure port
+		"--insecure-port", "0",
+		// authenticate kubelet to us
+		"--kubelet-certificate-authority", "/etc/homeworld/authorities/kubernetes.pem",
+		// authenticate us to kubelet
+		"--kubelet-client-certificate", "/etc/homeworld/keys/kubernetes-master.pem", "--kubelet-client-key", "/etc/homeworld/keys/kubernetes-master.key",
```
cryslith: shouldn't these paths use the constants?

platform/kubernetes/launch.go
```diff
@@ -0,0 +1,269 @@
...
+		return errors.Wrap(err, "while running scheduler")
+	}
+	return nil
+}
+
+func LaunchService(name string) error {
+	switch name {
+	case "apiserver":
+		return LaunchAPIServer()
+	case "ctrlmgr":
```
cryslith: Can this be `controller-manager` to match the name of the kubernetes command?

platform/kubernetes/wrapper/wrap.go
```diff
@@ -0,0 +1,85 @@
+package wrapper
```
cryslith: What is the purpose of this program?

platform/kubernetes/launch.go
```diff
@@ -14,6 +15,20 @@ import (
 	"github.com/sipb/homeworld/platform/kubernetes/wrapper"
 )
 
+func getVerbosityArgument() string {
+	verbosityBytes, err := ioutil.ReadFile("/etc/homeworld/config/verbosity")
```
cryslith: please document this somewhere

platform/setup-queue/deploy.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	err = wrapper.GenerateKubeConfigToFile(paths.KubernetesWorkerKey, paths.KubernetesWorkerCert, KubeConfigPath)
+	if err != nil {
+		return err
+	}
+	for _, file := range files {
+		name := path.Join(dir, file.Name())
+
+		log.Printf("deploying %s\n", name)
+		err := DeploySingle(KubeConfigPath, name)
```
cryslith: could spawn a goroutine and have everything in the queue run concurrently

platform/kube-state-metrics/launch.go
```diff
@@ -0,0 +1,41 @@
...
+	"strconv"
+)
+
+const KubeConfigPath = "/etc/homeworld/config/kubeconfig-kube-state-metrics"
+const MetricsPort = 9104
+const MetaMetricsPort = 9105
+
+func LaunchKSM(kubeconfig string, port, metaport int) error {
+	portStr := strconv.FormatInt(int64(port), 10)
+	metaportStr := strconv.FormatInt(int64(metaport), 10)
```
cryslith: port numbers are u16

platform/keysystem/worldconfig/keyserver.go
```diff
@@ -204,24 +204,34 @@ func GrantsForNodeAccount(c *config.Context, conf *SpireSetup, groups Groups, au
...
+		grants[SignKubernetesSupervisorAPI] = account.NewTLSGrantPrivilege(
+			auth.Kubernetes, false, 30*OneDay, "supervisor:"+node.Hostname,
+			nil,
+			// TODO: reduce the permissions granted to the supervisor nodes once the cluster has been configured
+			[]string{
+				"system:masters",
+			},
+		)
+	}
+
```
cryslith: make these an if-else pair instead of two if statements

platform/user-grant/grant.go
```diff
@@ -166,11 +173,71 @@ func (c *Config) CertGen(key crypto.PublicKey, user string) ([]byte, error) {
 	return certutil.FinishCertificate(certTemplate, c.IssuerCert, key, c.IssuerKey)
 }
 
+func (c *Config) CreateNamespace(namespace string) error {
```
cryslith: Can we add at least one check that just makes sure that `namespace` starts with `user-`?  I know that as written, this code only ever does just that, but this is a pretty critical invariant.

platform/spire/src/template.py
```diff
@@ -40,3 +41,42 @@ def template(filename_or_contents, keys, load=True) -> str:
 def template_all(filename_or_contents, keys_iterable, load=True) -> list:
     templ = Template(filename_or_contents, load=load)
     return [templ.template(keys) for keys in keys_iterable]
+
+
+def string_substitute(text: str, keys: dict) -> str:
```
cryslith: Suggestion: just use ordinary python format string syntax; this function can be defined as:
```
def string_substitute(text, keys):
    return text.format(**keys)
```

platform/spire/resources/setup.yaml
```diff
@@ -5,6 +5,7 @@ cluster:
   kerberos-realm: ATHENA.MIT.EDU
   mirror: debian.csail.mit.edu/debian
   user-grant-domain: homeworld.mit.edu
+  user-grant-email-domain: MIT.EDU
```
cryslith: There's one more copy of `setup.yaml` in the `deploy-chroot` directory (should be moved, but that's where it is right now).  Can you update that one too?

design/rbac.md
```diff
@@ -0,0 +1,75 @@
...
+-------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+system:node:<hostname>         | system:node    | The kubelet running on each master or worker node.    | Only the hostname, not the full domain name.
+supervisor:<hostname>          | system:masters | The setup-queue on the supervisor.                    | This is also used by the kube-state-metrics and prometheus services on the supervisor.
+apiserver:<hostname>           |                | The apiservers for their server certificates.         | This is never used to authenticate to the cluster, only to secure TLS connections to the apiservers.
+root:<principal>               | system:masters | Root admins when authenticating via the keysystem.    | The full principal, `user/root@ATHENA.MIT.EDU`, is used.
+root:direct                    | system:masters | Root admins when authenticating via keysystem bypass. | The prefix `root:` does not grant access; only the `system:masters` group does.
```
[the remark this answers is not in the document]
celskeggs: We don't get to change this. It's the name assigned by Kubernetes upstream.

platform/keysystem/keylocalcert/keylocalcert.go
```diff
@@ -44,6 +45,10 @@ func main() {
 	if err != nil {
 		logger.Fatal(err)
 	}
+	var organizations []string
+	if len(os.Args[7]) > 0 {
+		organizations = strings.Split(os.Args[7], ",")
+	}
```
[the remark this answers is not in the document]
celskeggs: I'm under the impression that `strings.Split` returns `[]string{""}` and not `[]string{}` when given an empty argument, which I think means I need these guards.

platform/upload/BUILD.bazel
```diff
@@ -76,28 +76,35 @@ genrule(
...
     name = "keyexport",
     outs = ["keyring.gpg"],
-    srcs = ["KEYID"],
-    cmd = "gpg --export -- \"$$(cat '$<')\" >'$@'",
+    srcs = [
+        "KEYID",
+        "SEQID",
+    ],
+    cmd = "gpg --export -- \"$$(cat '$(location :KEYID)')\" >'$@'",
     visibility = ["//visibility:public"],
```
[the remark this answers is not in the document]
celskeggs: I don't know of any way to make bazel re-run something each time, and I think part of the point of bazel is that you can't do that easily anyway.

platform/kubernetes/launch.go
celskeggs: Looking at the history, it comes from the very earliest versions of Hyades back in 2017, got contradicted in meaning when allow-privileged changed to true without changing the comment, and now has no real meaning. I'm going to eliminate it.

platform/kubernetes/wrapper/wrap.go
celskeggs: This library is used by various launcher programs to generate the kubeconfig needed to connect to the cluster, or to grab other information relating to launching and wrapping kubernetes or other components.

platform/keysystem/keylocalcert/keylocalcert.go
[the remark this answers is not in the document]
cryslith: You are right.  [Here is my excuse.](https://play.golang.org/p/_-eygUXarFI)

platform/setup-queue/deploy.go
celskeggs: Not sure why that'd improve things? I don't think shaving a few seconds off launch gets us much, in comparison to all the other optimizations we could do, and it adds unnecessary jitter w/r/t which one is deployed in which order.

platform/kube-state-metrics/launch.go
celskeggs: FormatInt takes an int64, so I have to convert to this.

platform/spire/src/template.py
celskeggs: Yeah, I'm really not sure why I didn't just go with that.

platform/keysystem/keylocalcert/keylocalcert.go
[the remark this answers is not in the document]
celskeggs: Hah, nice.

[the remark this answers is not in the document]
celskeggs: Heh nice.

platform/kubernetes/wrapper/wrap.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+	kvs := map[string]string{}
+	for _, line := range strings.Split(string(conf), "\n") {
+		line = strings.TrimSpace(line)
+		if len(line) > 0 && line[0] != '#' {
+			kv := strings.SplitN(line, "=", 2)
+			if len(kv) != 2 {
+				return nil, errors.New("incorrectly formatted cluster.conf")
```
cryslith: This error message should indicate the actual filename instead of assuming it was cluster.conf

Also - where do these configuration files come from and why this configuration format?

cryslith:
other than this 1 comment - appr

platform/kubernetes/wrapper/wrap.go
celskeggs: These configuration files are generated by a combination of spire and the keysystem. They're in this format so that they can be source'd by bash scripts, though this commit does eliminate most of those uses.

They should probably be converted to YAML at some point, once we finish eliminating bash launch scripts. I've filed #431 to remind us of this.

Issue #337, opened 2018-10-27
Disks installed incorrectly on real hardware


celskeggs:
On real hardware, the installer is actually on a USB stick, not on a CDROM. In practice, the USB drive shows up as sda, and the actual disk shows up as sdb.

This causes a few problems:
 - grub cannot install normally (... did we hard-code the path for it or something?), and must be installed manually
 - the disk numbering changes on reboot without the USB stick, which means that sdb no longer exists, and everything gets stuck in the initramfs on boot. (This can be fixed manually, by mounting the disk from the initramfs, and then regenerating it.)
 - the normal automatic partitioning settings are, for an unknown reason, skipped, letting the user fall back to the normal debian partitioning menu.

Given that we need to add Ceph support soon, it would be prudent for us to have a better disk-handling story.

ichung:
Some notes from our attempts to recreate the situation on toastfs-dev's VMs (i.e. putting the install media as an (emulated) USB drive instead of a cdrom)
- The bootloader only ever boots one "hard disk" at a time, which makes it annoying to automate the setup
- For some reason, the installer cannot detect the target disk in this configuration

celskeggs:
The reason that installation isn't fully automated when using a USB drive is that the code only autoselects a disk when there's only one inserted:

```
# See if any autopartition disks have been set
disks=""
if db_get partman-auto/disk && [ "$RET" ]; then
        disks="$RET"
fi

# If there's only one disk, then preseeding partman-auto/disk is
# unnecessary, and sometimes inconvenient in heterogeneous environments
if [ "$method" ] && [ -z "$disks" ]; then
        DEVS="$(get_auto_disks)"
        if [ "$(echo "$DEVS" | wc -l)" -eq 1 ]; then
                disks="$(cat "${DEVS%$TAB*}"/device)"
        fi
fi
```

This comes from the https://salsa.debian.org/installer-team/d-i source code under `packages/partman-auto/display.d/initial_auto`.

In our case, we have the choice to either hardcode the disk we use, patch the source code to autoselect the disk, or figure out some sort of step to automatically select it.

celskeggs:
Relevantly, the only reason a CDROM isn't treated as a valid disk is that there's special-case code for it:

```
get_auto_disks() {
        local dev device dmtype

        for dev in $DEVICES/*; do
                [ -d "$dev" ] || continue

                device=$(cat $dev/device)
                
[...]

Issue #340, opened 2018-10-31
Support VLAN trunking in network configuration


celskeggs:
The new server environment involves network ports that use VLAN trunking, which means we need to reconfigure how we set up server networks.

Here's what a manually patched /etc/network/interfaces file looks like:

```
# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

source /etc/network/interfaces.d/*

# The loopback network interface
auto lo
iface lo inet loopback

# The primary network interface
allow-hotplug eno1
iface eno1 inet manual

auto eno1.240
iface eno1.240 inet static
	address 18.85.240.10/20
	gateway 18.85.240.1
	# dns-* options are implemented by the resolvconf package, if installed
	dns-nameservers 18.70.0.160 18.71.0.151 18.72.0.3
	dns-search unassigned-domain

auto eno1.241
iface eno1.241 inet static
	address 192.168.240.110/24
```

VLAN 240 is the publicly routable VLAN. VLAN 241 is the IPMI VLAN, so that block is only included on the supervisor node.

We should add appropriate configuration options for VLAN trunking in the cluster configuration file -- we want to both be able to control whether VLAN trunking is used and which VLAN numbers are used.

Eventually, we should make sure that servers can have this configured heterogenously.

celskeggs:
Next step on this is to figure out how to get the Debian installer to use VLANs.

celskeggs:
Twelve years after VLAN support in d-i was first requested, there's still no support. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=433568

celskeggs:
Fixed via #440 and tested on real hardware.

PR #435, opened 2019-11-15
Fix USB drive installer handling

### LINKED ISSUE #338: Mismatch between USB and CDROM installation in dev and prod
In our development clusters, we insert our installer as an emulated CDROM. But in production, this is actually a USB drive. This means that issues like #332 and #337 don't show up in development, only in production, which is problematic.

We should make our development environments closer to the real thing by emulating a USB drive instead of a CDROM.

While we're at it, we should find a way to have disks show up as /dev/sda in test as well, rather than as /dev/vda or similar.

This would involve fixing both of our development virtualization methods:

- Update the libvirt XML configurations on toastfs-dev (under /root/hyades/)
- Update the qemu parameters generated in homeworld-admin-tools's virt.py

celskeggs:
I encountered trouble when I last tried to install on real hardware due to using a USB install drive instead of a CDROM. (We had been simulating a CDROM installer in our VM tests.)

This commit fixes this by:
 - Changing the autoinstaller to emulate the case where the installer is a USB hard drive, which reveals the problem in our integration tests.
 - Adding code to automatically select which hard drive to use for installation, ignoring USB drives.
 - Adding code to work around an issue in the debian installer that prevents grub from using the UUID of the main hard drive when booting.

See #337, which may be fixed by this. Fixes #338.

---

Checklist:

 - [x] ~I have written or updated appropriate documentation to cover this change.~ No documentation is required.

### DIFF
--- a/platform/spire/src/virt.py
+++ b/platform/spire/src/virt.py
@@ -261,11 +261,12 @@ class DebugContext:
 
 
 class VirtualMachine:
-    def __init__(self, node, tc: TerminationContext, cd=None, cpus=12, memory=2500):
+    def __init__(self, node, tc: TerminationContext, cd=None, cpus=12, memory=2500, cdrom_install=False):
         self.node = node
         self.cd = cd
         self.cpus = cpus
         self.memory = memory
+        self.cdrom_install = cdrom_install
         self.netif = get_node_tap(node)
         self.tc = tc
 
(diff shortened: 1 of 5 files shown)


cryslith:
Have you tested this on real hardware?

platform/spire/src/virt.py
```diff
@@ -288,12 +288,14 @@ def generate_qemu_args(self):
...
         if self.cd is None:
+            args += ["-drive", "media=disk,file=" + self.hd]
             args += ["-boot", "c"]
         else:
-            args += ["-cdrom", self.cd]
-            args += ["-boot", "d"]
+            # TODO: maybe set this readonly
+            args += ["-drive", "media=disk,format=raw,file=" + self.cd]
+            args += ["-drive", "media=disk,file=" + self.hd]
+            args += ["-boot", "c"]
```
cryslith: Could we leave it configurable whether to put the installer as CD-ROM or hard disk?

celskeggs:
I have not tested this on real hardware, because we still need #430 before we can do that. Is that a problem?

platform/spire/src/virt.py
celskeggs: We could, but do we think that's useful?

cryslith:
> I have not tested this on real hardware, because we still need #430 before we can do that. Is that a problem?

#430 is merged.  But the point was let's not close #337 until we verify that this actually fixes the problem

platform/spire/src/virt.py
cryslith: I don't see it as out of the question that someone might want to use an actual CD-ROM for install media.

celskeggs:
Woops, I meant #340, not #430. But sure, I'll remove the 'fixes' tag.

celskeggs:
@cryslith re-review?

celskeggs:
Merged with cryslith's previous code review approval; GitHub dismissed it because I rebased, but I made no changes beyond the rebase, so I considered it acceptable to merge without another code review cycle.

PR #437, opened 2019-11-15
Linting

### LINKED ISSUE #175: Linting
Do we want to lint our code? Do we want to integrate linting into our build process? I was thinking of pylint and https://github.com/koalaman/shellcheck specifically.

cryslith:
Add CI steps to run `pyflakes3` and `shellcheck`.

Closes #175.

---

Checklist:


cryslith:
Oh hmm, there is no `pyflakes3` package in ubuntu trusty...

tools/shellcheck-all.sh
```diff
@@ -0,0 +1,17 @@
...
+cd "$(git rev-parse --show-toplevel)"
+
+failed=""
+while IFS="" read -r file
+do
+    if grep -Eq '^#!(.*/|.*env +)(sh|bash|ksh)' "$file" || (echo "$file" | grep -Eq '.*.(ba)?sh$')
+    then
+        if ! shellcheck "$file"
+        then
+            failed=failed
```
celskeggs: Any reason we can't just put an `exit 1` here, instead of having this variable? (My bash is a bit rusty, so I suspect the answer is something like "this part technically runs in a subshell" or something.)

tools/shellcheck-all.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
+
+failed=""
+while IFS="" read -r file
+do
+    if grep -Eq '^#!(.*/|.*env +)(sh|bash|ksh)' "$file" || (echo "$file" | grep -Eq '.*.(ba)?sh$')
```
celskeggs: I believe the `echo | grep` here can be replaced with a usage of the regex matching operator?

tools/extract-apt-setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e -u
+
+cd "$(git rev-parse --show-toplevel)"
```
celskeggs: Can you make this commit (5e14b9649f2db81261fe0e570ca1a5a14c260b96) have a more detailed commit message? Currently, it doesn't explain the rationale for moving (which is obvious at this point but would be worth clarifying with a single sentence), and implies that extract-apt-setup is unchanged, when you've actually made some changes to the script.

.circleci/config.yml
```diff
@@ -7,10 +7,13 @@ jobs:
...
       - run:
           name: Install build tools
-          command: sudo apt-get update && sudo apt-get install build-essential cpio squashfs-tools debootstrap realpath pyflakes3
+          command: sudo apt-get update && sudo apt-get install build-essential cpio squashfs-tools debootstrap realpath pyflakes3 shellcheck
       - run:
           name: Run pyflakes3
           command: pyflakes3 .
+      - run:
+          name: Run shellcheck
+          command: tools/shellcheck-all.sh
```
celskeggs: Any reason this (850f63f41f9eef64bfa1c9345b5053ca318bd63e) isn't part of the last commit (f3d45368324947d523d65f33ad2948b33b969586)?

build-chroot/bump-pinned-bazel.sh
```diff
@@ -6,6 +6,7 @@ set -e -u
 BASEPATH="$(realpath "$(dirname "$0")")"
 
 WORKDIR=$(mktemp -d)
+# shellcheck disable=SC2064
```
celskeggs: Can we clarify in the commit message of 8a3bd45 that we're also just silencing some errors, and not just fixing them?

tools/shellcheck-all.sh
```diff
@@ -4,14 +4,14 @@ set -euo pipefail
 cd "$(git rev-parse --show-toplevel)"
 
 failed=""
-while IFS="" read -r file
+while IFS="" read -r -d '' file
 do
     if grep -Eq '^#!(.*/|.*env +)(sh|bash|ksh)' "$file" || (echo "$file" | grep -Eq '.*.(ba)?sh$')
     then
-        if ! shellcheck "$file"
+        if ! shellcheck -e SC1091,SC2015,SC2016 "$file"
```
celskeggs: Can we add a comment here explaining why we're disabling these shellcheck rules?

tools/shellcheck-all.sh
cryslith: This is so that you see all errors, rather than only being able to see errors in later files after you've fixed earlier ones.

tools/shellcheck-all.sh
cryslith: TIL

tools/extract-apt-setup.sh
cryslith: The changes to the script are strictly necessary to make it work from its new location.

.circleci/config.yml
cryslith: Because I intended to order the "fix shellcheck errors" step before this one

build-chroot/bump-pinned-bazel.sh
cryslith: Worth noting we only silence false positives - that is, the code is actually written in the optimal way and shellcheck is just overzealous in certain instances.

tools/extract-apt-setup.sh
cryslith:
```
tools: Move extract-apt-setup and debian-checksum to tools directory

This directory contains miscellaneous scripts which are useful for development.
- Rewrite extract-apt-setup to behave correctly in its new location
```

build-chroot/bump-pinned-bazel.sh
cryslith: Actually I say that, but it's not entirely true.  However, I'll leave a comment in the source code for that situation.

celskeggs: Sorry, yes, I mean -- can we mention in the commit message that some of what the commit does is silencing false positives?

.circleci/config.yml
```diff
@@ -2,15 +2,18 @@ version: 2
 jobs:
   build:
     machine:
-      image: circleci/classic:latest
+      image: ubuntu-1604:201903-01
```
celskeggs: Should we be concerned about moving from a `:latest` image to a particular version of an image?

cryslith: No.  [The circleci documentation](https://circleci.com/docs/2.0/configuration-reference/#available-machine-images) makes clear that `ubuntu-1604:201903-01` is the recommended image to use.

build-chroot/bump-pinned-bazel.sh
cryslith:
```
shell-scripts: Fix existing shellcheck errors 

Also silence some false positives, as well as a valid lint which depends on a separate issue.
```

celskeggs:
@cryslith ready to merge pending rebase

.circleci/config.yml
cryslith: And indeed I did, even before the rebase.  I still think whatever interface you're using to view the commits is not giving you them in topological order.

PR #440, opened 2019-11-16
Support for VLAN trunk upstreams

### LINKED ISSUE #227: Add monitoring and verification of systemd services
We should notice if they stop running, and our installation process should use these metrics to confirm that they're actually running.

I think these metrics are provided by the prometheus node exporter, but we'll still need scripts to  consume them.

This is a deferred part of #180.

### LINKED ISSUE #340: Support VLAN trunking in network configuration

celskeggs:
This change adds support to the autoinstaller to configure its network with VLAN trunking connectivity, fixes the installer to support this configuration, and turns this configuration on by default for testing.

Fixes #227, because I wanted that to help debug a broken Jenkins build. (Though the issue didn't end up reappearing, so /shrug.) Intended to fix issue #340, but that needs testing.

---

Checklist:

### DIFF
--- a/.jenkins/deploy-setup.yaml
+++ b/.jenkins/deploy-setup.yaml
@@ -7,6 +7,8 @@ cluster:
   user-grant-domain: usergrant.private
   user-grant-email-domain: EXAMPLE.ORG
 
+vlan: 612
+
 addresses:
   cidr-nodes: 10.1.0.0/24
   cidr-pods: 172.18.0.0/16
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -1,0 +1,234 @@
+# setup.yaml documentation
+
+This file documents the format of the `setup.yaml` file used by Homeworld to configure a Homeworld cluster.
+
+## High-Level Overview
+
+The `setup.yaml` file is a document formatted in YAML, containing a set of sections that have different meanings. You
+can refer to [the sample setup.yaml file](../platform/spire/resources/setup.yaml) to get a sense of what it looks like.
+
+These sections are:
+
+ * `cluster`: configuration of the important names in the cluster, such as the domain names and mirrors to use.
+ * `vlan`: configuration of trunked VLANs.
+ * `addresses`: configuration of the address ranges to use for the cluster.
+ * `dns-upstreams`: configuration of the DNS servers.
+ * `dns-bootstrap`: manual configuration of temporary DNS entries.
+ * `root-admins`: configuration of administrative access.
+ * `nodes`: a list of the physical nodes within the cluster.
+
+## Cluster Names
+
+Sample section:
+
+    cluster:
+      external-domain: mit.edu
+      internal-domain: hyades.local
+      etcd-token: <unique-token>
+      kerberos-realm: ATHENA.MIT.EDU
+      mirror: debian.csail.mit.edu/debian
+      user-grant-domain: homeworld.mit.edu
+      user-grant-email-domain: MIT.EDU
+
+Descriptions:
+
+ * `external_domain`: the publicly-addressible domain under which the nodes of the cluster can be resolved.
+
+   If this is, for example, `mit.edu`, a node named `rhombi` would be addressible under the hostname `rhombi.mit.edu`.
+
+   Note that these DNS entries do not necessarily need to exist, but it'll be easier if they do.
+
+ * `internal_domain`: the internal domain under which internal DNS addresses should be placed.
+
+   For example, if this is `hyades.local`, then any domain `<service>.<namespace>.svc.hyades.local` would resolve to the
+   service IP of the service `<service>` within the namespace `<namespace>` within the cluster. These domains should not
+   be publicly addressible, because the IP addresses that they refer to will not be publicly addressible either.
+
+ * `etcd-token`: an opaque token describing your particular cluster, to be used to identify individual etcd clusters.
+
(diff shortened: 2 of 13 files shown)


.jenkins/deploy-setup.yaml
```diff
@@ -7,6 +7,8 @@ cluster:
   user-grant-domain: usergrant.private
   user-grant-email-domain: EXAMPLE.ORG
 
+vlan: 0
+
```
cryslith: let's make the `vlan` key optional instead of "0 for no vlan"

platform/spire/src/virt.py
```diff
@@ -18,10 +18,17 @@
...
     return "spirebr%s" % ip.packed.hex().upper()
 
 
+def get_bridge_access(ip):
+    """The name to be used for the vlan interface providing host-side access to the VMs."""
+    return "spireac%s" % ip.packed.hex().upper()
+
+
 def get_node_tap(node):
+    """The name to be used for the tap interface providing network to a particular QEMU VM."""
```
cryslith: I think I'm a little lost here.  Can you explain what the effect of this network configuration is?

platform/spire/resources/servicemap.yaml
```diff
@@ -0,0 +1,41 @@
+services:
```
cryslith: What ensures that this file doesn't get out of sync with the actual set of services we run?

docs/cluster-configuration.md
```diff
@@ -0,0 +1,233 @@
...
+
+   A supervisor node does not need to be up for the regular operations of the cluster, but will need to be up at least
+   intermittently to allow key renewal to occur on the other nodes.
+
+ * `master`: a node that runs the management services of the Kubernetes cluster. These nodes do not run user-provided
+   code, but do run a full Kubernetes node stack, in addition to the management services.
+
+ * `worker`: a node that runs user containers for the Kubernetes cluster.
+
+The number of master nodes should generally be one more than a power of two, because they will need quorum for their
```
cryslith: Shouldn't this just be an odd number rather than 1 + power of 2?  What am I missing here?

celskeggs: Yeah, I should really have just said that. I don't know why I didn't.

platform/spire/resources/servicemap.yaml
celskeggs: The entries in the file won't get out of sync, because any deviation from the actual set of services that run will make things fail CI. However, nothing will catch us if we add additional services and forget to add them to this file, and I'm not sure exactly how we'd do that. Any ideas?

platform/spire/src/virt.py
celskeggs: In the previous model, we had a bridge on the host and a bunch of tap interfaces associated with the bridge. Each VM would be connected to a tap interface by QEMU. We'd configure an IP address on the bridge so that we could access the VM network, and set up forwarding so that the VMs on the network could access the public internet through the host.

We don't need any change to the bridge or the tap interfaces to support passing VLAN trunk traffic between the VMs; we just need to make sure that the host's bridge IP address is on a VLAN instead of on the bare bridge. To do that, we need to add a subinterface on the bridge (not an interface attached to the bridge, but a VLAN interface on the bridge) that uses the correct VLAN number and bind our IP address to that instead of the bridge directly.

This way, the VMs can't access the network with the host as gateway unless they set up VLAN interfaces.

So the tap interfaces function exactly the same as they did before, and we just add an extra vlan interface to use for binding instead of the bridge when VLAN trunking is enabled.

celskeggs:
I had to upgrade the validation mechanism for setup.yaml in order to easily support having optional keys, but now "no VLAN trunking" is configured by not including the `vlan: ` key. I've also fixed the "odd number" thing.

platform/spire/resources/servicemap.yaml
cryslith: Yes, I was talking about the latter scenario of course.  I don't have a good idea of how to solve this right now but we should keep it in mind

celskeggs:
Both checks and review passed when I tried to merge... but this PR was set to be merged into 337-disk-install instead of master, so I had to rewind master, change the base, and re-merge this PR to get it to be acknowledged as merged -- and so the checks and review were dismissed.

Issue #163, opened 2017-12-18
spire should have a command for re-deploying the cluster configuration


celskeggs:
This means deleting all the instances of cluster.conf and local.conf, and then rolling all of the keyclients.

celskeggs:
Resolved via #441.

PR #441, opened 2019-11-17
163 add config redeploy

AngelAlvie:
Spire does not have a convenient way to redeploy a cluster configuration to a running cluster. This change adds a command `spire seq redeploy` which takes the current cluster configuration from `setup.yaml` and redeploys it to a running cluster. This change simply pushes the new config to each of the cluster nodes, and restarts their respective keyclient (or in the case of the supervisor, keyserver) services.

This change also allows the keyserver to notify systemd when it is up, using the same mechanism as is used by the keyclient. In working on this change, I have noticed that restarting the keyclient may take more than five minutes, simply because the keyclient fails to receive the new config from the keyserver as a result of the keyserver not being up before we restart the keyclients. As a result the keyclient would fail and wait 5 minutes to restart. Allowing the keyserver to notify systemd when it is up again means that spire can sequence the operations needed to redeploy the config without waiting an arbitrary amount of time, or pinging the keyserver to see if the keyserver is responding to http.

See #163.

---

Checklist:

 - [ ] I have written or updated appropriate documentation to cover this change.

### DIFF
--- a/platform/keysystem/keyserver/keyserver.go
+++ b/platform/keysystem/keyserver/keyserver.go
@@ -16,13 +16,11 @@ func main() {
 	_, onstop, err := keyapi.Run(":20557", logger)
 	if err != nil {
 		logger.Fatal(err)
-	} else {
-		err := exec.Command("systemd-notify", "--ready").Run()
-		if err != nil {
-			logger.Fatal("failed to notify systemd of readiness: %v\n", err)
-		} else {
-			// service is up, wait for kill signal
-			logger.Fatal(<-onstop)
-		}
 	}
+	err = exec.Command("systemd-notify", "--ready").Run()
+	if err != nil {
+		logger.Fatal("failed to notify systemd of readiness: %v\n", err)
+	}
+	// service is up, wait for kill signal
+	logger.Fatal(<-onstop)
 }
--- a/platform/keysystem/keyserver/keyserver.go
+++ b/platform/keysystem/keyserver/keyserver.go
@@ -3,6 +3,7 @@ package main
 import (
 	"log"
 	"os"
+	"os/exec"
 
 	"github.com/sipb/homeworld/platform/keysystem/keyserver/keyapi"
 )
@@ -16,6 +17,12 @@ func main() {
 	if err != nil {
 		logger.Fatal(err)
 	} else {
-		logger.Fatal(<-onstop)
+		err := exec.Command("systemd-notify", "--ready").Run()
+		if err != nil {
+			logger.Fatal("failed to notify systemd of readiness: %v\n", err)
+		} else {
+			// service is up, wait for kill signal
+			logger.Fatal(<-onstop)
+		}
 	}
 }
--- a/platform/spire/src/seq.py
+++ b/platform/spire/src/seq.py
@@ -60,6 +60,14 @@ def sequence_supervisor(ops: command.Operations, skip_verify_keygateway: bool=Fa
         if node.kind == 'supervisor':
             ops.add_subcommand(infra.infra_sync, node.hostname)
 
+@command.wrapseq
+def sequence_redeploy_config(ops: setup.Operations) -> None:
+    "redeploy a cluster configuration to a running cluster"
+    # push new config to the keyserver
+    ops.add_subcommand(setup.redeploy_keyserver)
+    # push new config to each keyclient and restart
+    ops.add_subcommand(setup.redeploy_keyclients)
+
 
 class IterativeVerifier(command.Simple):
     def __init__(self, verifier, max_time, pause=2.0):
@@ -122,4 +130,5 @@ main_command = command.SeqMux("commands about running large sequences of cluster
     "ssh": sequence_ssh,
     "supervisor": sequence_supervisor,
     "cluster": sequence_cluster,
+    "redeploy": sequence_redeploy_config,
 })
--- a/platform/spire/src/seq.py
+++ b/platform/spire/src/seq.py
@@ -61,12 +61,12 @@ def sequence_supervisor(ops: command.Operations, skip_verify_keygateway: bool=Fa
             ops.add_subcommand(infra.infra_sync, node.hostname)
 
 @command.wrapseq
-def sequence_redeploy_config(ops: setup.Operations) -> None:
+def sequence_redeploy_config(ops: command.Operations) -> None:
     "redeploy a cluster configuration to a running cluster"
     # push new config to the keyserver
-    ops.add_subcommand(setup.redeploy_keyserver)
+    setup.redeploy_keyserver(ops)
     # push new config to each keyclient and restart
-    ops.add_subcommand(setup.redeploy_keyclients)
+    setup.redeploy_keyclients(ops)
 
 
 class IterativeVerifier(command.Simple):
--- a/platform/spire/src/setup.py
+++ b/platform/spire/src/setup.py
@@ -85,8 +85,6 @@ def redeploy_keyserver(ops: Operations) -> None:
                             configuration.Config.get_setup_path(), CONFIG_DIR + "/setup.yaml")
         # restart the keyserver
         ops.ssh("restart keyserver on @HOST", node, "systemctl", "restart", "keyserver.service")
-        # TODO: eliminate sleep by only finishing previous command only when service is restarted.
-        ops.ssh("waiting for keyserver to come back up on @HOST", node, "sleep", "5")
 
 def redeploy_keyclients(ops: Operations) -> None:
     config = configuration.get_config()
--- a/platform/spire/src/setup.py
+++ b/platform/spire/src/setup.py
@@ -70,31 +70,32 @@ def setup_keyserver(ops: command.Operations) -> None:
         ssh_cmd(ops, "enable keyserver on @HOST", node, "systemctl", "enable", "keyserver.service")
         ssh_cmd(ops, "start keyserver on @HOST", node, "systemctl", "restart", "keyserver.service")
 
-def redeploy_keyserver(ops: Operations) -> None:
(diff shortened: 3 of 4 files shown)


platform/keysystem/keyserver/keyserver.go
```diff
@@ -16,6 +17,12 @@ func main() {
 	if err != nil {
 		logger.Fatal(err)
 	} else {
```
cryslith: Let's remove the unnecessary `else` clauses after `if ... { logger.Fatal() }` to avoid arbitrarily deep nesting

platform/spire/src/seq.py
```diff
@@ -129,4 +134,5 @@ def configure(command: list, parser: argparse.ArgumentParser):
     "supervisor": wrapseq("set up and verify functionality of entire supervisor node (keysystem + ssh)",
                                sequence_supervisor),
     "cluster": wrapseq("set up and verify kubernetes infrastructure operation", sequence_cluster),
+    "redeploy": wrapseq("redeploy a cluster configuration to a running cluster", sequence_redeploy_config),
```
cryslith: needs rebase

platform/spire/src/setup.py
```diff
@@ -141,6 +145,31 @@ def setup_keyserver(ops: Operations) -> None:
...
+    for node in config.nodes:
+        if node.kind != "supervisor":
+            continue
+        # delete the existing configs
+        ops.ssh_rm("delete existing cluster config from @HOST", node, STATICS_DIR + "/cluster.conf")
+        ops.ssh_rm("delete existing keyserver config from @HOST", node, CONFIG_DIR + "/setup.yaml")
+        # redeploy new config
+        ops.ssh_upload_bytes("reupload cluster config to @HOST", node,
+            configuration.get_cluster_conf().encode(), STATICS_DIR + "/cluster.conf")
+        ops.ssh_upload_path("upload cluster setup to @HOST", node,
```
cryslith: any reason for one message to say "reupload" vs "upload"?

platform/spire/src/setup.py
```diff
@@ -108,6 +108,10 @@ def ssh_mkdir(self, name: str, node: configuration.Node, *paths: str, with_paren
         options = ["-p"] if with_parents else []
         self.ssh(name, node, "mkdir", *(options + ["--"] + list(paths)))
 
+    def ssh_rm(self, name: str, node: configuration.Node, path:str, allow_recursive: bool=False) -> None:
+        options = ["-rf"] if allow_recursive else ["-f"]
+        self.ssh(name, node, "rm",  *(options + ["--"] + [path]))
```
celskeggs: Is there a reason we need `allow_recursive` as an argument? I don't see it used anywhere.

platform/spire/src/setup.py
```diff
@@ -141,6 +145,31 @@ def setup_keyserver(ops: Operations) -> None:
...
+                            configuration.Config.get_setup_path(), CONFIG_DIR + "/setup.yaml")
+        # restart the keyserver
+        ops.ssh("restart keyserver on @HOST", node, "systemctl", "restart", "keyserver.service")
+
+def redeploy_keyclients(ops: Operations) -> None:
+    config = configuration.get_config()
+    for node in config.nodes:
+        # delete existing cluster configuration from non-supervisor nodes
+        if node.kind != "supervisor":
+            ops.ssh_rm("delete existing cluster config from @HOST", node, CONFIG_DIR + "/cluster.conf")
```
celskeggs: Can you explain in comments why we don't delete the configs from the supervisor?

platform/spire/src/seq.py
AngelAlvie: I have rebased this branch and have adjusted the changes to fit the new version of spire. I have also tested the changes to ensure that they still work.

platform/spire/src/setup.py
AngelAlvie: There was no reason that it should have said "upload", I have changed the offending line to "reupload". Good catch!

platform/spire/src/setup.py
AngelAlvie: We do not need the `allow_recursive` argument right now. I added it since I figured it may be useful later, but for now, I have removed it in the rewrite of this change after the rebase.

platform/spire/src/setup.py
AngelAlvie: I have changed the line `# delete existing cluster configuration from non-supervisor nodes` to 
```python
        # we do not delete the config on the supervisor because the supervisor
        # client config and keyserver config are in the same location
```

To reflect why we don't delete the config from the supervisor. As you mentioned earlier, the keyserver and keyclient probably should be using the same config directory, maybe this is something we should change here too? (Also i might rewrite this comment again, perhaps the language
```python
        # do not delete the cluster.conf on the supervisor because the supervisor
        # keyclient and keyserver use the same cluster.conf
        # TODO(alvareza): modify keyclient and keyserver to use different config directories
```
would work?

platform/keysystem/keyserver/keyserver.go
AngelAlvie: I have removed the unnecessary else clause. I suppose I didn't realize that `logger.Fatal` actually calls exit, hence why I wrote it that way.

celskeggs:
Just as a note, please rebase, rather than merging from master. It's easier to review PRs if there aren't merge commits, and it keeps the history cleaner. (I will review this later when I have time -- maybe next meeting.)

platform/spire/src/setup.py
celskeggs: I'd file an issue, rather than adding a TODO to the codebase, especially because it's not really a TODO to change something in this code exactly.

platform/keysystem/keyserver/keyserver.go
```diff
@@ -15,7 +16,12 @@ func main() {
 	_, onstop, err := keyapi.Run(":20557", logger)
 	if err != nil {
 		logger.Fatal(err)
+	}
+	err = exec.Command("systemd-notify", "--ready").Run()
+	if err != nil {
+		logger.Fatal("failed to notify systemd of readiness: %v\n", err)
 	} else {
```
cryslith: mind removing the other one too, while you're at it?

platform/spire/src/setup.py
```diff
@@ -91,6 +94,37 @@ def admit_keyserver(ops: command.Operations) -> None:
...
+    config = configuration.get_config()
+    for node in config.nodes:
+        # we do not delete the config on the supervisor because the supervisor
+        # client config and keyserver config are in the same location
+        if node.kind != "supervisor":
+            ssh_rm(ops, "delete existing cluster config from @HOST", node, CONFIG_DIR + "/cluster.conf")
+        ssh_rm(ops, "delete existing local config from @HOST", node, CONFIG_DIR + "/local.conf")
+        # restart local keyclient (will regenerate configs on restart)
+        ssh_cmd(ops, "restart keyclient daemon on @HOST", node, "systemctl", "restart", "keyclient.service")
+
```
cryslith: Do we want to ability to redeploy these individually?  If so then they need to be placed into the subcommand hierarchy somewhere

AngelAlvie: To clarify, you mean to say that we may want to allow redeploy keyclients and redeploy keyserver as the subcommands and not redeploy each server individually?

If so I think it would be fine to support the sub commands, although, I'm not sure how often one may want to redeploy the config to the keyserver without doing so also to the keyclients.

cryslith: I was just wondering if you possibly meant to add subcommands for these functions and forgot; I have no opinion on it otherwise.

platform/spire/src/setup.py
AngelAlvie: I have filed a new issue to track this issue, see #475. Furthermore, I added the comment (minus the todo) in 2f980ef4e2116cb750b6cb5e765bed381c4eab68

platform/keysystem/keyserver/keyserver.go
AngelAlvie: extraneous `else` clause removed in 6ae8ae9bc86c6781a41dfb28b72d9c6ad29d97d4.

platform/spire/src/setup.py
AngelAlvie: Okay, I didn't mean to add them as sub-commands, and I don't think that separating the redeploy step into two phases will be necessary. I have decided to keep them under one command, but thanks for letting me know!

platform/spire/src/seq.py
```diff
@@ -59,6 +59,14 @@ def sequence_supervisor(ops: command.Operations, skip_verify_keygateway: bool=Fa
         if node.kind == 'supervisor':
             ops.add_subcommand(infra.infra_sync, node.hostname)
 
+@command.wrapseq
+def sequence_redeploy_config(ops: command.Operations) -> None:
+    "redeploy a cluster configuration to a running cluster"
+    # push new config to the keyserver
+    setup.redeploy_keyserver(ops)
+    # push new config to each keyclient and restart
+    setup.redeploy_keyclients(ops)
```
celskeggs: Since the `seq` file is mostly for things that are composites of other commands, I think the "correct" thing to do is either:
 - to change redeploy_keyserver and redeploy_keyclients to be commands and use add_subcommand
 - OR to move this command to setup.py
But I think this PR has been held up long enough, and it's not like spire is very consistent as things are, so I don't think it's worth making this change.

celskeggs:
@AngelAlvie you did another merge commit, FYI.

celskeggs:
oh, and FYI this build is going to break until we merge #478, because we need the bump to the debian ISO version.

celskeggs:
@AngelAlvie good to go; you can rebase now.

celskeggs:
Nice job, @AngelAlvie!

PR #450, opened 2019-11-28
build/deploy misc fixes

### LINKED ISSUE #422: deploy chroot machine names conflict
@AngelAlvie ran into an issue where he couldn't run one of the deploy chroot scripts, because it had the same machine name as the deploy chroot container that another user was running.

We should set machine names to different values when using systemd-nspawn, so that this sort of conflict doesn't happen.

Note that this also means modifying /etc/hosts to use the new hostname.

cryslith:
Fix a couple of minor issues in the build/deploy process.

Fixes #422.

---

Checklist:


docs/build.md
```diff
@@ -11,6 +11,7 @@ Install the following on your host system:
  * sudo
  * systemd-container
  * gpg
+ * curl
```
celskeggs: Can you add a small note to the commit message to explain why curl is being added? i.e. "We depended on curl, but never specified it."
(This is to make it clear that the commit isn't supposed to mean "We need to start depending on curl for [reason].")

celskeggs:
Didn't wait for CI because CI had passed earlier and only a commit message changed.

PR #464, opened 2019-12-15
Upgrade build chroot and debian containers to buster

### LINKED ISSUE #428: Update to Debian 10 (Buster)
We're currently built around Stretch. It shouldn't be too hard to upgrade to Buster. Stretch isn't EOL yet, but it will be at some point, and we'll get nice shiny newer versions of things on Buster.

Elements of this:
 - [ ] Upgrade build chroot to buster
 - [ ] Upgrade container images to buster
 - [ ] Upgrade deploy chroot to buster
 - [ ] Upgrade installation CD to buster

celskeggs:
This PR upgrades us to the latest version of Debian, including working around [bug 946783](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=946783) in debootstrap.

Note that the deploy chroot does not have an automated test, and that I'm not intending to add one in this PR.

Fixes #428.

---

Checklist:


platform/debian/clean_fakechroot.py
```diff
@@ -23,3 +23,7 @@
             continue
         os.remove(path)
         os.symlink(os.path.join("/", rootrel), path)
+
+if os.path.islink(os.path.join(rootfs, "proc")):
```
krawthekrow: Should add a comment or todo in the code as well.

celskeggs:
Current status, so that I remember:

I'm currently running into a problem where the newer qemu available on buster objects to having the same CD inserted into multiple VMs simultaneously:

    qemu-system-x86_64: Initialization of device ide-hd failed: Failed to get "write" lock
    Is another process using the image [/cluster/cluster-1125.iso]?

This is reasonable, because they really should be read-only, but when I try to set them read-only, I get another error:

    qemu-system-x86_64: Initialization of device ide-hd failed: Block node is read-only

docs/build.md
```diff
@@ -20,14 +20,67 @@ Set up your build chroot:
     $ export HOMEWORLD_CHROOT="$HOME/homeworld-chroot"     # this can be any directory you choose
     $ ./build-chroot/create.sh
 
-You might consider adding the variable declaration to your ~/.profile.
+You might consider adding the variable declaration to your ~/.bashrc.
```
cryslith: Not everyone uses bash.  `.profile` is the correct place for environment variables.

celskeggs:
It looks like it's impossible to set `ide-hd` disks to be read-only, because it's hardcoded to only accept read-only options from `ide-cd`. I think this means that we can't set it read-only. However, we can turn off the file locking. It's not a perfect solution, but it's better than nothing.

celskeggs:
@krawthekrow I believe this is ready for review whenever you have the chance

cryslith:
> It looks like it's impossible to set `ide-hd` disks to be read-only, because it's hardcoded to only accept read-only options from `ide-cd`. I think this means that we can't set it read-only. However, we can turn off the file locking. It's not a perfect solution, but it's better than nothing.

Can we see if we can use the `snapshot` option to avoid writing back to the original disk image?  That ought to let us avoid disabling file locking

tools/debian-checksum.sh
```diff
@@ -1,25 +1,34 @@
...
-curl -sO "http://debian.csail.mit.edu/debian/dists/stretch/main/installer-amd64/$VERSION/images/SHA256SUMS"
+curl -sO "http://debian.csail.mit.edu/debian/dists/${RELEASE}/main/installer-amd64/${VERSION}/images/SHA256SUMS"
 
 echo 'verifying SHA256SUMS against Release' >&2
 # [tail] because the sha256's are after the md5's :P
 echo "$(grep "main/installer-amd64/$VERSION/images/SHA256SUMS" Release | tail -n 1 | cut -d " " -f 2)" SHA256SUMS | sha256sum --check --strict >&2
 
-grep './netboot/mini.iso' SHA256SUMS | cut -d " " -f 1
+echo
+echo "new version block:"
```
cryslith: Can we maintain the separation between stdout and stderr in this script?  Same for the message above

tools/debian-checksum.sh
```diff
@@ -1,25 +1,34 @@
...
+
+echo
+echo "VERSION = \"${VERSION}\""
+echo "RELEASE = \"${RELEASE}\""
+echo "HASH = \"$(grep './netboot/mini.iso' SHA256SUMS | cut -d " " -f 1)\""
```
cryslith: Let's make clear that `HASH` means the hash of `netboot/mini.iso`.  e.g. `MINI_ISO_HASH`.

build-chroot/debootstrap-1.0.106.deb.sha256
```diff
@@ -0,0 +1 @@
+14da318a90c66b9ef4be97b950bffd418033598c44f12f472d3501edec103f40  debootstrap-1.0.106.deb
```
cryslith: Can we compute this hash in `debian-checksum.sh`?

platform/spire/src/virt.py
```diff
@@ -339,7 +352,8 @@ def boot_with_io(self, phase, output_callback=None, text: bytes=None, delay=None
                 output = log_output
 
             if text:
-                stdin = delayed_yielder(delay, text)
+                assert delay is not None and chardelay is not None
```
cryslith: this is not what `assert` is for; see #303

platform/spire/src/virt.py
```diff
@@ -325,10 +338,10 @@ def generate_qemu_args(self):
         return args
 
     def check_module_loaded(self):
-        # TODO: don't blindly load kvm-intel; check type of system first
-        sudo("modprobe", "kvm", "kvm-intel")
+        # TODO: don't blindly load kvm_intel; check type of system first
+        modprobe("kvm", "kvm_intel")
 
-    def boot_with_io(self, phase, output_callback=None, text: bytes=None, delay=None):
+    def boot_with_io(self, phase, output_callback=None, text: bytes=None, delay=None, chardelay=None):
```
cryslith: This is getting really hacky.  Maybe we should try using [QMP](https://wiki.qemu.org/QMP) for this, which is designed for scripting.

tools/debian-checksum.sh
```diff
@@ -1,25 +1,34 @@
...
 echo 'verifying SHA256SUMS against Release' >&2
 # [tail] because the sha256's are after the md5's :P
 echo "$(grep "main/installer-amd64/$VERSION/images/SHA256SUMS" Release | tail -n 1 | cut -d " " -f 2)" SHA256SUMS | sha256sum --check --strict >&2
 
-grep './netboot/mini.iso' SHA256SUMS | cut -d " " -f 1
+echo
+echo "new version block:"
+
+echo
+echo "VERSION = \"${VERSION}\""
```
cryslith: nitpick: Let's replace `\"` with `'` here, since this is expected to be placed into another script.

platform/spire/src/virt.py
celskeggs: yeah, this is very much a "sanity-check my own code" assert

platform/spire/src/virt.py
celskeggs: I suspect this may actually be a problem with the guest, not the host. But yeah, we should consider using a better tool at some point. I'm hoping to not have that hold up this PR, though.

tools/debian-checksum.sh
celskeggs: Hmm okay

build-chroot/debootstrap-1.0.106.deb.sha256
celskeggs: There's no point in doing so; we can't update this until the bug gets fixed, and at that point, we won't need the pin anymore.

tools/debian-checksum.sh
cryslith: I think you might have gone too far now - now everything goes to stderr.  I was thinking the `VERSION`, `RELEASE`, and `MINI_ISO_HASH` lines should go to stdout and everything else should go to stderr

celskeggs: Not certain I see why we would care for this, since it's a tool current designed for manual use by a developer, but sure.

cryslith: The way I was using it before was by piping it into my text editor when I wanted to use it :)

celskeggs: oh yeah that'd be smart

celskeggs:
Merged according to @cryslith's approval of the non-rebased version.

PR #478, opened 2020-02-01
Upgrade bazel to 2.1.0

### LINKED ISSUE #444: Upgrade to latest Bazel version
We're currently on Bazel 0.26.1. That's more than five months out of date.

The following non-point releases have been released since the version we're using:
 - [ ] 0.27
 - [ ] 0.28
 - [ ] 0.29
 - [ ] 1.0
 - [ ] 1.1
 - [ ] 1.2
 - [ ] 2.0
 - [ ] 2.1

This isn't urgent, because we pin bazel now, but should be done at some point in the near future for the sake of not having trouble next time we upgrade all of our dependencies.

celskeggs:
Upgrade us to a newer version of bazel. Fix incompatibilities along the way.

Also updates the debian-iso version and adds a helper script for that, because that got broken along the way.

Fixes #444.

---

Checklist:


tools/update-defs.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python3
```
cryslith: This implementation seems overcomplicated for such a simple string-processing task.

platform/.bazelrc
```diff
@@ -1,2 +1,2 @@
 startup --host_jvm_args=-Xmx4G
-build --workspace_status_command bazel/workspace-status.sh
+build --workspace_status_command bazel/workspace-status.sh --host_force_python=PY2
```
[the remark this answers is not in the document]
celskeggs: #489!

tools/update-defs.py
celskeggs: Does this shorter one seem more appropriate? I don't want to make it too simple, because I want to make sure it will error out rather than mangle the generated file or silently drop data.

celskeggs:
Merged per previously passed checks and review on 382fa0f before rebase.

PR #479, opened 2020-02-03
 deploy-chroot: Create kvm device instead of bind-mounting it.

### LINKED ISSUE #476: Clobbering of /dev/kvm permissions across deploy chroots on RHOMBI
When two people try to use deploy chroots at the same time, each of them will reassign the ownership of /dev/kvm to be root:kvm, so that users in the kvm (like the automatically-created one) will be able to run kvm-accelerated virtual machines.

Unfortunately, different deploy chroots may have different group IDs for the kvm group, especially if they have different operating system version numbers. This means they start to play "tug-of-war" over the ownership; whichever one doesn't have the right ownership will fail to launch VMs, and when the deploy chroot is restarted to fix it, it takes away the right ownership from the other deploy chroot.

This also has the side concern that it may change access on the host to a group that it shouldn't be giving access to -- like a group used to isolate vulnerable system services. While this isn't much of a concern on RHOMBI, which doesn't run very many services, it could be a problem in other deployment situations.

The reason they need to have the same ownership is that each chroot has a bind mount of /dev/kvm. We should just mknod (or maybe even cp) the node into each chroot, which will allow each deploy chroot to set its own ownership appropriately for its own device.

cryslith:
Fixes #476.

---

Checklist:

 - [x] ~I have formatted any Go code that I have changed with gofmt.~
 - [x] ~I have written or updated appropriate documentation to cover this change.~
 - [x] ~I have confirmed that this change is covered by at least one appropriate test run by CI.~

### DIFF
--- a/deploy-chroot/enter-internal.sh
+++ b/deploy-chroot/enter-internal.sh
@@ -3,6 +3,11 @@
 set -euo pipefail
 
 sudo mount -o rw,remount /proc/sys
-sudo chgrp kvm /dev/kvm
+if grep -wq kvm /proc/misc; then
+    sudo mknod --mode='ug=rw,o=' /dev/kvm c 10 "$(grep -w kvm /proc/misc | cut -f 1 -d ' ')"
+    sudo chgrp kvm /dev/kvm
+else
+    echo 'note: no kvm support; not creating /dev/kvm' >&2
+fi
 cd /cluster
 exec ssh-agent bash
--- a/deploy-chroot/enter-internal.sh
+++ b/deploy-chroot/enter-internal.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -euo pipefail
+
+sudo mount -o rw,remount /proc/sys
+sudo chgrp kvm /dev/kvm
+cd /cluster
+exec ssh-agent bash
--- a/deploy-chroot/enter.sh
+++ b/deploy-chroot/enter.sh
@@ -10,5 +10,21 @@ then
 fi
 
 cd "$(dirname "$0")"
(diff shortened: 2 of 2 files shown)


deploy-chroot/enter-internal.sh
```diff
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 sudo mount -o rw,remount /proc/sys
+sudo mknod --mode='ug=rw,o=' /dev/kvm c 10 "$(grep -w kvm /proc/misc | cut -f 1 -d ' ')"
```
celskeggs: How is this going to behave if kvm support happens to be missing?

cryslith: poorly.

deploy-chroot/enter-internal.sh
```diff
@@ -3,7 +3,9 @@
 set -euo pipefail
 
 sudo mount -o rw,remount /proc/sys
-sudo mknod --mode='ug=rw,o=' /dev/kvm c 10 "$(grep -w kvm /proc/misc | cut -f 1 -d ' ')"
+if grep -wq kvm /proc/misc; then
+    sudo mknod --mode='ug=rw,o=' /dev/kvm c 10 "$(grep -w kvm /proc/misc | cut -f 1 -d ' ')"
+fi
 sudo chgrp kvm /dev/kvm
```
celskeggs: Maybe stick the chgrp in the if?
Might also be worth explicitly warning that it couldn't create the device node?

cryslith: wooops

PR #486, opened 2020-02-09
Add metallb load balancer.

### LINKED ISSUE #361: Set up stop-gap load balancer
We already have an internal load-balancing system, provided by kube-proxy, and the real advantage of something like GLB (#336) is that it can scale horizontally, but we don't actually need horizontal scaling of our load balancing at this point.

So really we just need to be able to bind a public IP address on the subnet and forward it to a kube-proxy endpoint. And that would get us everything we currently need. We should set up something simple and wait until later to actually set up a full load balancer.

cryslith:
This adds the metallb load balancer in layer 2 mode.  Note that this doesn't provide any sort of high-availability.  There's also no changes to `spire virt`, so to use it with that, the external IP ranges need to be from the same subnet that spire configures on the host interface (i.e. `cidr-nodes`).

Closes #361.

---
todo:
- [ ] automated testing
- [ ] figure out what to do about SecurityContextDeny
- [ ] resplit commits

---

Checklist:

 - [ ] I have split up this change into one or more appropriately-delineated commits.
 - [ ] The first line of each commit is of the form "[component]: do something"
 - [ ] I have written a complete, multi-line commit message for each commit.
 - [ ] I have formatted any Go code that I have changed with gofmt.
 - [ ] I have written or updated appropriate documentation to cover this change.
 - [ ] I have confirmed that this change is covered by at least one appropriate test run by CI.
 - [ ] If my change includes new or modified functionality, I have tested that the changes work as expected.
 - [ ] I have assigned this issue to an appropriate reviewer. (Choose @celskeggs if you are not otherwise certain.)


platform/spire/src/verify.py
```diff
@@ -353,6 +353,13 @@ def check_user_grant():
 
     print("autogenerated rolebinding for user", repr(authority.UPSTREAM_USER_NAME), "passed basic check!")
 
+@command.wrap
+def check_metallb():
+    "verify that metallb is running"
+    expect_prometheus_query_exact('sum(up{job="metallb-controller"})', 1, "metallb controller is online")
+    expect_prometheus_query_exact('sum(up{job="metallb-speaker"})', 1, "metallb speaker is online")
+    print("metallb is running")
+
```
celskeggs, line 362: I would really like an automated check that the metallb services aren't just running, but also can handle requests to the exposed IP addresses. This might be too much work to include in this PR, but we should at least file an issue for it.

(This would be similar to our components like dns-monitor and pull-monitor.)

celskeggs:
Very nice work!

platform/upload/BUILD.bazel
```diff
@@ -11,6 +11,9 @@ ocis = {
     "flannel-monitor": "//flannel-monitor:oci",
     "kube-dns-main": "//kube-dns:kube-dns-main",
     "kube-dns-sidecar": "//kube-dns:kube-dns-sidecar",
+    "metallb-controller": "//metallb:controller",
+    "metallb-speaker": "//metallb:speaker",
+    "metallb-mirror": "//metallb:mirror",
```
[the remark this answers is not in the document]
celskeggs: I see your change to fix this; I notice that `         "mirror_digest": ":mirror.ocidigest", ` means that this OCI will probably still be built, which I don't think we need, since `mirror_digest` doesn't appear to be used anywhere.

Further, do we even need the build defined for the mirror OCI anymore, if we don't need to build it?

platform/spire/src/verify.py
cryslith: Yep, I will file a ticket

PR #487, opened 2020-02-10
Eliminate zip_file rule

### LINKED ISSUE #405: Eliminate zip_file from codebase
~~We're currently maintaining our own version of zip_file.bzl borrowed from an old version of Google's Nomulus repo; they are no longer maintaining this file, because they switched to Bazel, and it was never split out into a separate repository. This means a maintenance load on us to support zip files, as seen in #404.~~

~~Luckily, there's already an attempt to support a `pkg_zip` rule in rules_pkg: https://github.com/bazelbuild/rules_pkg/pull/65. Once this gets merged, we should switch to using that instead. (And if that PR gets abandoned, we should consider whether we have the capacity to help get it finished and merged, because it'd be great to not have to maintain this ourselves.)~~

We no longer need to use zip_file; #487 is a demonstration of how we can eliminate this dependency.

celskeggs:
The zip_file rule is a lot of work to maintain, and the only thing we're doing with it could be replaced with built-in bazel functionality ... at least, as of recently. Let's get rid of it.

Fixes #405.

---

Checklist:


platform/spire/src/command.py
```diff
@@ -315,7 +315,12 @@ def __call__(self, depth=0, dry_run=False):
 
 
 def main_invoke(command):
-    parser = argparse.ArgumentParser(description="Administrative toolkit for deploying and maintaining Hyades clusters")
+    # we have to hardcode the name "spire" here, because using bazel's py_binary tool for packaging our program into a
+    # zip file means we'll get the name of our __main__.py, rather than the name of the original command invoked.
```
cryslith: should we file a bug against rules_python for this?

platform/spire/BUILD.bazel
```diff
@@ -1,43 +1,30 @@
...
+        "//spire/resources:resourcelib",
+        "//upload:uploadlib",
+        "//user-grant:kubelib",
+        "//version:versionlib",
+    ],
+    main = "__main__.py",
+)
 
-zip_file(
+filegroup(
```
cryslith: do we need this intermediate filegroup, or can we just reference `:spirepy` directly?

platform/spire/src/configuration.py
```diff
@@ -265,8 +265,8 @@ def get_kube_spec_vars(extra_kvs: dict=None) -> dict:
     return kvs
 
 
-def get_single_kube_spec(name: str, extra_kvs: dict=None) -> str:
-    templ = resource.get_resource("clustered/%s" % name).decode()
+def get_single_kube_spec(path: str, extra_kvs: dict=None) -> str:
+    templ = resource.package_resource(path).decode()
```
cryslith: I really don't like how all uses of these files now need to know whether to use `spire.resources.resource` or `resource.package_resource`; and the latter has bazel's weird `//` thing going on.  Is there any way we can improve this?

platform/spire/BUILD.bazel
```diff
@@ -1,43 +1,30 @@
...
-mappings.update({
-    "__main__/" + ref.lstrip("/").replace(":", "/"): "resources/clustered/" + name
-    for name, ref in clustered.items()
-})
+py_binary(
+    name = "spirepy",
+    srcs = glob([
+        "src/*.py",
+    ]),
+    data = [
```
cryslith: what exactly does `py_binary` do with `data`?  It's not well documented

platform/python/template_init.py
```diff
@@ -0,0 +1,10 @@
+# this is a template file that gets instantiated for any bazel package with python-accessible resources.
+# it provides a convenience function, but its simple presence as __init__.py is enough to denote a package's existence.
```
cryslith: I'm not sure I understand the resulting python package structure, can you explain it?

platform/spire/BUILD.bazel
celskeggs: We do. py_binary normally doesn't produce an executable zip file as output, so we have to force it to by specifying a filegroup with the output_group option.

platform/spire/src/command.py
celskeggs: I'm not sure there's much they can do about it? They have to pass a particular name to python in order to get the right script to run. I don't think they can pass the original name without it just not working -- unless they add another layer of stub script that runs within the target python interpreter, but that might mess up other semantics.

We could file an issue on upstream or on our own repo so that we remember this later, but I don't think this is a real problem for us. If we ever rename spire, we can just change this piece of code to reflect the new name.

platform/spire/BUILD.bazel
celskeggs: So this is what py_binary builds into the zip file:

```
$ unzip -l bazel-bin/spire/spirepy.zip
Archive:  bazel-bin/spire/spirepy.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
    13135  2010-01-01 00:00   __main__.py
        0  2010-01-01 00:00   __init__.py
        0  2010-01-01 00:00   runfiles/__main__/__init__.py
        0  2010-01-01 00:00   runfiles/__main__/spire/src/__init__.py
        0  2010-01-01 00:00   runfiles/__main__/spire/__init__.py
      950  2010-01-01 00:00   runfiles/__main__/spire/src/__main__.py
    13162  2010-01-01 00:00   runfiles/__main__/spire/src/access.py
     6802  2010-01-01 00:00   runfiles/__main__/spire/src/authority.py
    11888  2010-01-01 00:00   runfiles/__main__/spire/src/command.py
    10181  2010-01-01 00:00   runfiles/__main__/spire/src/configuration.py
     3471  2010-01-01 00:00   runfiles/__main__/spire/src/deploy.py
     3040  2010-01-01 00:00   runfiles/__main__/spire/src/infra.py
     6479  2010-01-01 00:00   runfiles/__main__/spire/src/iso.py
     1792  2010-01-01 00:00   runfiles/__main__/spire/src/keycrypt.py
     5903  2010-01-01 00:00   runfiles/__main__/spire/src/keys.py
      653  2010-01-01 00:00   runfiles/__main__/spire/src/metadata.py
     5622  2010-01-01 00:00   runfiles/__main__/spire/src/packages.py
     1426  2010-01-01 00:00   runfiles/__main__/spire/src/query.py
      588  2010-01-01 00:00   runfiles/__main__/spire/src/resource.py
     4896  2010-01-01 00:00   runfiles/__main__/spire/src/seq.py
    11587  2010-01-01 00:00   runfiles/__main__/spire/src/setup.py
     1779  2010-01-01 00:00   runfiles/__main__/spire/src/ssh.py
     2140  2010-01-01 00:00   runfiles/__main__/spire/src/template.py
      998  2010-01-01 00:00   runfiles/__main__/spire/src/util.py
    17096  2010-01-01 00:00   runfiles/__main__/spire/src/verify.py
    22667  2010-01-01 00:00   runfiles/__main__/spire/src/virt.py
      433  2010-01-01 00:00   runfiles/__main__/dns-monitor/__init__.py
     1240  2010-01-01 00:00   runfiles/__main__/dns-monitor/kubernetes.yaml
      433  2010-01-01 00:00   runfiles/__main__/dnsmasq/__init__.py
     2853  2010-01-01 00:00   runfiles/__main__/dnsmasq/kubernetes.yaml
      433  2010-01-01 00:00   runfiles/__main__/flannel/__init__.py
     2209  2010-01-01 00:00   runfiles/__main__/flannel/kubernetes.yaml
      433  2010-01-01 00:00   runfiles/__main__/flannel-monitor/__init__.py
     4608  2010-01-01 00:00   runfiles/__main__/flannel-monitor/kubernetes.yaml
      433  2010-01-01 00:00   runfiles/__main__/spire/resources/__init__.py
      202  2010-01-01 00:00   runfiles/__main__/spire/resources/isolinux.cfg.serial
      417  2010-01-01 00:00   runfiles/__main__/spire/resources/kubeconfig-local.yaml
     1737  2010-01-01 00:00   runfiles/__main__/spire/resources/netcfg.postinst
     3716  2010-01-01 00:00   runfiles/__main__/spire/resources/postinstall.sh
      826  2010-01-01 00:00   runfiles/__main__/spire/resources/prepartition.sh
    18755  2010-01-01 00:00   runfiles/__main__/spire/resources/preseed.cfg.in
     3949  2010-01-01 00:00   runfiles/__main__/spire/resources/prometheus.yaml
     1097  2010-01-01 00:00   runfiles/__main__/spire/resources/servicemap.yaml
      831  2010-01-01 00:00   runfiles/__main__/spire/resources/setup.yaml
     1611  2010-01-01 00:00   runfiles/__main__/spire/resources/setup-schema.yaml
     1865  2010-01-01 00:00   runfiles/__main__/spire/resources/sshd_config
      433  2010-01-01 00:00   runfiles/__main__/upload/__init__.py
        7  2010-01-01 00:00   runfiles/__main__/upload/BRANCH_NAME
       47  2010-01-01 00:00   runfiles/__main__/upload/DOWNLOAD_URL
     1280  2010-01-01 00:00   runfiles/__main__/upload/keyring.gpg
      433  2010-01-01 00:00   runfiles/__main__/user-grant/__init__.py
     2726  2010-01-01 00:00   runfiles/__main__/user-grant/kubernetes.yaml
      433  2010-01-01 00:00   runfiles/__main__/version/__init__.py
       47  2010-01-01 00:00   runfiles/__main__/version/GIT_VERSION
     3561  2010-01-01 00:00   runfiles/bazel_tools/tools/python/py3wrapper.sh
---------                     -------
   203303                     55 files
```

Unlike our previous mechanism, we can't actually choose where each file goes in this output tree, except by moving them around in our source tree. The `data` parameter is part of calculating the transitive closure of dependencies that get included; everything we put here will end up in the final runfiles tree. So we use it to pull in all of the things like our `kubernetes.yaml` files and our `GIT_VERSION`/`BRANCH_NAME`/`DOWNLOAD_URL` files and our `spire/resources/*` files et cetera.

platform/spire/src/configuration.py
celskeggs: With py_binary, we no longer get to remap the paths at which our dependencies get stuck into the output zipfile. (See https://github.com/sipb/homeworld/pull/487#discussion_r379573921.) This means that we need to pull in our resources from locations based on where they were in the original source tree, instead of being able to stick them all into a single unified `resources` directory and pull them all from there.

`spire.resources.resource("x")` pulls in a file `x` that was stored in `//spire/resources`. I could have also done this for (most of) the others -- we could `import dnsmasq` and do `dnsmasq.resource("kubernetes.yaml")` instead of `resource.package_resource("//dnsmasq:kubernetes.yaml")`, but that seemed like it would be potentially more confusing.

We can't do that for the paths with dashes in them -- like `dns-monitor` -- because those aren't valid python package names. There's probably a workaround for that, too, either based in renaming those directories in our source tree or renaming them in another way, but that also seemed messy.

I chose to reuse the `//package:filename` syntax; we didn't have to do that. It could be `resource.package_resource("dnsmasq", "kubernetes.yaml")` instead, but this way it seemed like it would be very clear that it was referencing a bazel package, and you'd know exactly where to look for the kubernetes.yaml, instead of having to figure out that bazel packages were being mapped to python packages before you could understand the reference.

I agree that this is somewhat ugly; it is not clear to me that there are easy improvements to the situation, but I'm interested in your input.

Maybe it works better to switch _everything_ to bazel's `//` syntax, for consistency?

platform/python/template_init.py
celskeggs: py_binary produces a package structure that corresponds directly to the bazel source tree. See https://github.com/sipb/homeworld/pull/487#discussion_r379573921.

The way the PYTHONPATH is set up lets us access particular things in a few ways:

 - `import dnsmasq` imports the `//dnsmasq` python package (via its `__init__.py` file, which is therefore mandatory for the package to be usable by python)
 - similarly, `import spire.resources` refers to the `//spire/resources` python package.
 - we can also import python modules in the `//spire:src` directory, because that's where the source code comes from.
 - we would be able to import `__main__.spire.resources` as well, except that `__main__` is interpreted as a special package in python and cannot be used like that. but we could use this if we wanted to reference something built in another workspace, like `@ceph//...` or something.

So the idea of template_init.py, as instantiated by py_resources into many copies of `__init__.py`, is to make it possible for us to access resources from a variety of bazel packages by turning them into stub python packages. This means that `pkgutil.get_data` can work, because they can be imported. We could manually do something around figuring out where the runfiles directory is stored, and extracting files directly from it, but this seems somewhat cleaner?

platform/spire/BUILD.bazel
cryslith: Hmm, I can't find the documentation on that output group.

celskeggs: Bold of you to assume that there is any documentation.

(I learned about it by running across the right GitHub issue.)

platform/spire/src/configuration.py
cryslith: I think we should switch everything to a single syntax for consistency; I don't have an opinion on whether bazel's or two arguments is better.  I also think `resource.package_resource` is too long and it should just be `resource.get` or something

celskeggs: Try that.

celskeggs:
Merged per @cryslith's approval and CI check before rebase against master.

PR #493, opened 2020-02-21
deploy-chroot: fix systemd-nspawn on buster

celskeggs:
Something changed about how permissions work under Debian Buster, so now that our development machine is on Buster, we have to pass systemd-nspawn the `--keep-unit` parameter to make it able to run the mknod command from within the chroot.

It also appears that, without `--keep-unit`, we end up potentially using the same machine scope (such as deploy-chroot.scope) when we have similarly-named deploy chroot directories. Adding `--keep-unit` fixes this as well.

No filed issue is associated.

---

Checklist:


celskeggs:
Yes, I tried adding that exact capability, and it didn't help, hence this solution.

PR #496, opened 2020-02-28
spire: Use short version of command docstring in output.

### LINKED ISSUE #477: Only use first docstring line as command description in spire
We get the following output from the cluster setup process:

            -- set up and verify ssh access to the supervisor node --
              --
        request SSH access to the cluster
    
        no_add_to_agent: do not add the resulting ssh key to ssh-agent
         --

This is supposed to be:

            -- set up and verify ssh access to the supervisor node --
              -- request SSH access to the cluster --

This is because the docstring on the access_ssh function looks like this:

```
@command.wrap
def access_ssh(no_add_to_agent: bool=False):
    """
    request SSH access to the cluster

    no_add_to_agent: do not add the resulting ssh key to ssh-agent
    """
```

This is a perfectly reasonable use of a docstring. But we shouldn't print the whole thing -- just the first actual line.

cryslith:
Fixes #477 

---

Checklist:

 - ~I have written or updated appropriate documentation to cover this change.~
 - ~I have confirmed that this change is covered by at least one appropriate test run by CI.~

### DIFF
--- a/platform/spire/src/command.py
+++ b/platform/spire/src/command.py
@@ -43,17 +43,18 @@ class Mux:
         subparsers = parser.add_subparsers()
 
         for component, subcommand in self.mapping.items():
-            doc = inspect.getdoc(subcommand)
-            short_doc = doc.split('\n')[0] if doc else None
             subparser = subparsers.add_parser(
                 component,
-                description=doc,
-                help=short_doc,
+                description=inspect.getdoc(subcommand),
+                help=subcommand.short_doc(),
                 formatter_class=argparse.RawDescriptionHelpFormatter)
             try:
                 subcommand.configure(command + [component], subparser)
             except AttributeError as e:
                 raise Exception("error configuring subcommand {!r}".format(subcommand)) from e
+
+    def short_doc(self):
+        return self.__doc__
 
 
 def add_dry_run_argument(parser: argparse.ArgumentParser, dest: str):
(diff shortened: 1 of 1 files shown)


platform/spire/src/command.py
```diff
@@ -43,18 +43,21 @@ def configure(self, command: list, parser: argparse.ArgumentParser):
         subparsers = parser.add_subparsers()
 
         for component, subcommand in self.mapping.items():
-            doc = inspect.getdoc(subcommand)
+            doc = subcommand.short_doc()
             short_doc = doc.split('\n')[0] if doc else None
```
celskeggs: Are these local variables used for anything anymore?

platform/spire/src/command.py
```diff
@@ -68,6 +71,9 @@ def configure(self, command: list, parser: argparse.ArgumentParser):
         add_dry_run_argument(parser, 'dry_run_outer')
         add_show_commands_argument(parser, 'show_commands_outer')
 
+    def short_doc(self):
+        return self.__doc__
+
```
celskeggs: Why do we have to override the superclass's implementation? Isn't this just the same implementation as Mux?

platform/spire/src/command.py
cryslith: oops

platform/spire/src/command.py
cryslith: oops

platform/spire/src/command.py
cryslith: on the bright side - pyflakes caught this in CI :)

Issue #29, opened 2017-07-01
Set up Ceph


celskeggs:
We've done nothing on this.

celskeggs:
@synergistics @Lchu613: You're now on the critical path, so:

 * What's the current status of this?
 * Can this be done by the end of DC5 (Oct 8)? If not, DC6 (Oct 22)?

celskeggs:
Since I haven't heard from either of you, this is getting deferred to DC6.

celskeggs:
As per separate conversations, I'm taking this.

celskeggs:
Ceph appears to very much not support encryption over the wire, which means we either need an overlay network, patches to Ceph, or to encrypt everything going into it.

1. Overlay networks will prevent use of block storage and CephFS outside of specially configured systems, which is bad.
2. Patching Ceph seems infeasible, given how much they've tried to implement this.
3. Encrypting everything going into Ceph with require using dm-crypt block overlays and object gateway server-side-encryption, and prevent use of CephFS.
4. Switching to an alternative to Ceph appears to be blocked by a lack of good alternatives to Ceph.

The best option appears to be entry 3 on the above list, which means that the Ceph cluster itself will be deployed as authenticated but unencrypted, and consumers will have to provide the encryption layer.

celskeggs:
For key management, it seems like this will involve a limited set of predefined static keys, which need to be generated in the keysystem and distributed by it.

According to analysis of the ceph-authtool, the output format is reasonably easy to generate:

    [client.admin]
    	key = AQBRVwVavY4SLhAANoj1BVGSSN/VbYrV9PFTxA==

These base64-encoded 28 bytes are split up as (in hex): `0100ssssssssnnnnnnnn1000kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk`

In these, `s` is the time in seconds, `n` is the time in nanoseconds, and `k` is a randomly-generated 16-byte key.

It might be easier to generate this format ourselves rather than try to invoke ceph-authtool.

ichung:
For documentation purposes: are those time fields little- or big-endian?

celskeggs:
Little-endian.

celskeggs:
As it turns out, it is essentially impossible to actually make Ceph's authentication secure, due to how certain keys are replicated over insecure channels.

We're looking into using [wireguard](https://www.wireguard.com/) to solve this, which would act like a higher-quality and easier-to-use version of IPSec for our purposes.

[...]

PR #498, opened 2020-02-29
[WIP] ceph: add basic build of binaries and OCI

celskeggs:
In order to meet our supply chain protection standards, we have to build our own version of Ceph to use with Rook. This commit brings in rules_foreign_cc, a set of bazel rules to more easily run cmake builds, and then uses those rules to build Ceph using cmake.

Of course, we end up needing to patch both Ceph and rules_foreign_cc to make them work correctly together, and we end up needing to install many more packages into the build chroot.

I am not yet planning on merging this PR, because I want to make more progress on Ceph before concluding that I did this part correctly, but I would like a code review as things stand.

The testing carried out to determine that this worked correctly was manually deploying a container running the `ceph` image, execing into it with kubectl, and confirming that every one of the ceph commands could be launched. (This took a lot of iteration, to make sure all of the dependencies -- both built as part of Ceph and not built as part of Ceph -- worked out.)

See #29.

---

Checklist:

 - I have written or updated appropriate documentation to cover this change.
 - [ ] I have confirmed that this change is covered by at least one appropriate test run by CI.


build-chroot/packages.list
```diff
@@ -1,3 +1,16 @@
+
+# ceph
+bc btrfs-progs cmake cpio cryptsetup-bin cython cython3 gdisk git gperf jq libaio-dev libbabeltrace-ctf-dev
+libbabeltrace-dev libblkid-dev libexpat1-dev libgoogle-perftools-dev libibverbs-dev libkeyutils-dev
+libldap2-dev libleveldb-dev liblttng-ust-dev libleveldb-dev liblttng-ust-dev libnss3-dev libsnappy-dev libssl-dev
+libtool libudev-dev libxml2-dev lsb-release parted pkg-config python python-all-dev python-cherrypy3 python-nose
+python-pecan python-prettytable python-setuptools python-sphinx python-werkzeug python3-all-dev python3-setuptools
+socat uuid-runtime virtualenv xfslibs-dev xfsprogs xmlstarlet yasm zlib1g-dev librabbitmq-dev libpython-dev
+libcunit1-dev liboath-dev libncurses-dev libcurl4-openssl-dev librdmacm-dev liblz4-dev
```
cryslith: are all of these actually being used still?  e.g. was virtualenv only used for the node thing we decided to cut?

platform/ceph/BUILD.bazel
```diff
@@ -0,0 +1,156 @@
...
+    name = "ceph",
+    binaries = BINARIES,
+    cache_entries = {
+        "WITH_MANPAGE": "off",
+        "WITH_PYTHON3": "off",
+        "WITH_LTTNG": "off",
+        "WITH_EMBEDDED": "off",
+        "WITH_TESTS": "off",
+        "WITH_CEPHFS": "off",
+        # using system boost shaves off about 25% of the build time in my tests
```
cryslith: out of curiosity, how long does this take to build?

platform/ceph/BUILD.bazel
```diff
@@ -0,0 +1,156 @@
...
+        # disabled because the dashboard requires npm, which requires nodeenv,
+        # which requires working pip in a virtualenv, which is not working for
+        # some reason relating to DNS resolution.
+        # (and it's too much hell to make it functional for me to want to deal
+        # with this mess right now.)
+        "WITH_MGR_DASHBOARD_FRONTEND": "off",
+    },
+    lib_source = "@ceph//:source",
+    make_commands = [
+        "make -j12",
```
cryslith: why 12?

platform/ceph/deps.bzl
```diff
@@ -0,0 +1,21 @@
...
+    rules_foreign_cc_dependencies()
+    http_archive(
+        name = "ceph",
+        url = "https://download.ceph.com/tarballs/ceph_" + VERSION + ".orig.tar.gz",
+        sha256 = SHA256,
+        patch_cmds = [
+            # remove symlinks that create cycles (which break glob)
+            "find -name '.qa' -delete",
+            # remove filenames with ":" in them (which are disallowed in filegroups)
+            "rm src/test/common/test_blkdev_sys_block/sys/dev/block/8:0 src/test/common/test_blkdev_sys_block/sys/dev/block/9:0",
```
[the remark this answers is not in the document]
celskeggs: There definitely might be some expected names that happen to include colons, but verifying that there aren't any further names like this is a great idea.

platform/ceph/deps_early.bzl
```diff
@@ -0,0 +1,19 @@
...
+        commit = "ed3db61a55c13da311d875460938c42ee8bbc2a5",
+        patches = [
+            # we need this patch so that bazel's -D__DATE__="redacted" CFLAG doesn't become
+            # -D__DATE__=redacted, which causes code that uses __DATE__ to break.
+            "//ceph:foreign_cc/0001-fix-date-quoting.patch",
+            # (see https://github.com/bazelbuild/rules_foreign_cc/issues/239
+            # and https://github.com/bazelbuild/rules_foreign_cc/pull/362)
+
+            # we need this so that we can correctly reference the generated libraries
+            "//ceph:foreign_cc/0002-more-output-groups.patch",
```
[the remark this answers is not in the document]
celskeggs: Maybe? Upstream hasn't been super responsive AFAICT, so I'm not excited about carefully filing a report.

build-chroot/packages.list
celskeggs: I believe I copy-pasted the things they wanted, rather than verifying each one. Do you think we need to worry about having unnecessary packages?

platform/ceph/BUILD.bazel
celskeggs: At the point where I wrote this, something like three hours.

platform/ceph/BUILD.bazel
celskeggs: Half the number of CPUs on RHOMBI, which allows for nice parallelism without completely stepping on everything bazel is doing.

But yeah, this should probably be scaled automatically based on # of CPUs. The real problem is that it's a really bad thing to be running two different job servers at once; but we don't have much of an option, because without parallelism, this step takes hours, and there's no current way to have make use bazel's build server.

platform/ceph/deps_early.bzl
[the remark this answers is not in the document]
celskeggs: Filed an issue with upstream: https://github.com/bazelbuild/rules_foreign_cc/issues/376

platform/ceph/BUILD.bazel
```diff
@@ -92,7 +92,22 @@ cmake_external(
...
+        # so we compute the number of processors available, and divide by two
+        # rounding up.
+        #
+        # if this ever accidentally becomes 'make -j' due to the expr commands
+        # failing, then it's going to be terrible and (from experience)
+        # probably hang the entire build machine for several minutes before the
+        # OOM killer gets fed up with Bazel, but I can't come up with a good
+        # way around that, and hopefully this code is correct enough that it
+        # never happens.
+        "make -j$(expr $(expr 1 + $(cat /proc/cpuinfo | cut -f 1 | grep -E '^processor$' | wc -l)) / 2)",
```
cryslith: - I don't think you need the nested `$(expr ...)` here, `expr` understands parentheses just fine.

- use `</proc/cpuinfo ...`, not `cat /proc/cpuinfo | ...`

- There are a couple ways to detect the inner command failing in bash, such as setting a variable and using an explicit test.  But at this level of complexity it might be appropriate to refactor this into a separate script.

celskeggs: - Will fix
 - Will fix
 - The make_commands parameter is badly implemented, and there appears to be no way to embed a reference to an external file in it, hence inlining the logic that really deserves a separate script. I will try to update the logic to be smarter, just inline.

platform/ceph/BUILD.bazel
```diff
@@ -0,0 +1,171 @@
...
+    "rados-classes/libcls_numops.so.1",
+    "rados-classes/libcls_otp.so.1",
+    "rados-classes/libcls_rbd.so.1",
+    "rados-classes/libcls_refcount.so.1",
+    "rados-classes/libcls_rgw.so.1",
+    "rados-classes/libcls_sdk.so.1",
+    "rados-classes/libcls_timeindex.so.1",
+    "rados-classes/libcls_user.so.1",
+    "rados-classes/libcls_version.so.1",
+]
```
cryslith, line 70: the lists of binaries and libraries could change over time.  Can we leave a comment explaining how you figured them out?  (Or not, if it isn't as simple as "go look at <some file> in the ceph source")

platform/ceph/BUILD.bazel
```diff
@@ -0,0 +1,171 @@
...
+        "WITH_CEPHFS": "off",
+        # using system boost shaves off about 25% of the build time in my tests
+        "WITH_SYSTEM_BOOST": "on",
+        "WITH_RADOSGW_BEAST_FRONTEND": "on",
+        "WITH_FUSE": "off",
+        # disabled because the dashboard requires npm, which requires nodeenv,
+        # which requires working pip in a virtualenv, which is not working for
+        # some reason relating to DNS resolution.
+        # (and it's too much hell to make it functional for me to want to deal
+        # with this mess right now.)
```
cryslith, line 90: I think it's a fine justification without this acerbic parenthetical :)

platform/ceph/BUILD.bazel
cryslith: in fact - do we know that we need all of these binaries?  would we improve build time by not linking so many of them?

celskeggs: I believe that the binaries are "all of the binaries that Ceph generates" and the libraries are "all of the libraries needed for those to work (tested empirically)", but I don't exactly remember. I don't think we know exactly which ones Rook requires, but I suspect the answer is "most of them."

PR #502, opened 2020-04-11
Update gazelle to v0.20.0

### LINKED ISSUE #501: Update gazelle
We're using an old version which lacks some nice features such as support for the new go module system.  This will be an issue when adding dependencies on go code which uses the new modules.

Since the new version supports go.mod and go.sum, I also think we should switch to autogenerating all our gazelle rules using `gazelle update-repos` once we do this.  This will make go dependency management require significantly less manual effort.

cryslith:
Also upgrade the prometheus golang client dependencies with it,
since that's what I needed the new gazelle version for in the first place.

Closes #501.

---

Checklist:

### DIFF
--- a/platform/cri-o/build.patch
+++ b/platform/cri-o/build.patch
@@ -150,7 +150,7 @@ index 3a0a989..9e0efbe 100644
      name = "package-srcs",
 --- a/vendor/golang.org/x/crypto/ed25519/BUILD.bazel	2019-09-29 15:57:46.980000000 -0400
 +++ b/vendor/golang.org/x/crypto/ed25519/BUILD.bazel	2019-09-29 15:36:54.862000000 -0400
-@@ -2,15 +2,9 @@
+@@ -2,12 +2,9 @@
  
  go_library(
      name = "go_default_library",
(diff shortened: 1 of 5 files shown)


platform/cri-o/build.patch
```diff
@@ -162,9 +162,5 @@ index 3a0a989..9e0efbe 100644
      importmap = "github.com/cri-o/cri-o/vendor/golang.org/x/crypto/ed25519",
      importpath = "golang.org/x/crypto/ed25519",
      visibility = ["//visibility:public"],
--    deps = [
--        "//vendor/crypto/ed25519:go_default_library",
--        "//vendor/golang.org/x/crypto/ed25519/internal/edwards25519:go_default_library",
--    ],
-+    deps = ["//vendor/golang.org/x/crypto/ed25519/internal/edwards25519:go_default_library"],
```
celskeggs, line 169: I'm curious why this change is part of this commit. Could this perhaps be explained in the commit message?

cryslith: Well, it seems that gazelle started doing what we want; I'm not sure why.  Note that my change replaces `change X to Y` with `leave Y unchanged`; i.e. gazelle has stopped incorrectly producing `X` and is now correctly producing `Y` instead, so we don't need to do anything.

celskeggs: Hmm, okay. That makes sense.

celskeggs:
Waiting on the last two checklist items to merge.

Issue #490, opened 2020-02-13
Upgrade Kubernetes to 1.16.7


krawthekrow:
This is not the latest version, but it might help with #458. More details there.

PR #514, opened 2020-04-24
Upgrade CRI-O to 1.16.3

krawthekrow:
See commit message. Unfortunately this does not seem to fix the memory leak, and the number of exit files still grows apparently unboundedly.

See #490 .

---

Checklist:

 - [ ] If my change includes new or modified functionality, I have tested that the changes work as expected.
 - [ ] I have assigned this issue to an appropriate reviewer. (Choose @celskeggs if you are not otherwise certain.)


platform/cri-o/conmon.patch
```diff
@@ -0,0 +1,307 @@
...
++# genrule(
++#     name = "config.h-genrule",
++#     outs = [":conf/config.h"],
++#     tools = ["//cmd/crio-config"],
++#     cmd = """
++#         REL="$$(realpath "$(location //cmd/crio-config)")"
++#         cd \"$(@D)\"
++#         "$${REL}"
++#     """,
++# )
```
celskeggs, line 18: Why are these blocks commented out instead of removed?

platform/cri-o/conmon.patch
```diff
@@ -0,0 +1,307 @@
...
++        "src/ctrl.c",
++        "src/ctrl.h",
++        "src/ctr_stdio.c",
++        "src/ctr_stdio.h",
++        "src/parent_pipe_fd.c",
++        "src/parent_pipe_fd.h",
++        "src/ctr_exit.c",
++        "src/ctr_exit.h",
++        "src/runtime_args.c",
++        "src/runtime_args.h",
```
celskeggs, line 63: This is a long list... maybe we could use a `glob` instead?

platform/cri-o/crio.conf.upstream
```diff
@@ -152,6 +153,7 @@ additional_devices = [
 
 # Path to OCI hooks directories for automatically executed hooks.
 hooks_dir = [
+	"/usr/share/containers/oci/hooks.d",
```
celskeggs, line 156: What do we put in this hooks directory?

platform/cri-o/deps.bzl
```diff
@@ -13,3 +14,14 @@ def cri_o_dependencies():
         ],
         patch_args = ["-p1"],
     )
+
+    go_repository_alt(
+        name = "com_github_containers_conmon",
+        commit = "1bddbf7051a973f4a4fecf06faa0c48e82f1e9e1", # 2.0.15
+        importpath = "github.com/containers/conmon",
+        build_file_generation = "off",
```
celskeggs, line 22: If we're turning off build file generation, what's the point of having this be a go_repository rule instead of a git_repository rule?

platform/cri-o/conmon.patch
krawthekrow: Oh whoops, I missed that completely. Thanks for catching that!

platform/cri-o/crio.conf.upstream
krawthekrow: That's the upstream file (to detect upstream config file generation changes). The array is empty in `crio.conf.in`.

platform/cri-o/deps.bzl
krawthekrow: The git_repository rule seems to cause the patched-in symlink to no longer be a symlink for some reason, and I can't figure out why. In any case, I'll change the go_repository_alt back into a go_repository.

celskeggs: Okay; we might want to do some digging on that at some point. If you don't want to do it now, can you file a low-priority issue?

platform/cri-o/crio.conf.upstream
celskeggs: Oh, yeah, misread that. :heavy_check_mark:

